PR #12, opened 2022-11-01
Remove PR title linting

p-schuler:
This was used to try to force the use of [conventional commit messages](https://www.conventionalcommits.org/en/v1.0.0/). It only enforces the PR title, but not the commit message itself. The flow is not clear and to avoid confusion, we want to disable this.


atifaziz (approved):
🎉

PR #13, opened 2022-11-01
Add registration watchdog

### LINKED ISSUE #9: Add registration watchdog
The registration watchdog keeps track of announcements from capability providers and evicts stale entries from the registry.

## Acceptance Criteria

- [x] There is a (global) configurable TTL for capability registrations.
- [x] A bad TTL isn't ignored.
- [x] TTL has a reasonable minimum.
- [x] Registrations have a timestamp for the last seen announcement message.
- [x] Expired registrations are removed.
- [x] Watchdog can be cancelled (e.g. graceful shutdown during <kbd>CTRL</kbd>+<kbd>C</kbd>)
- [x] The feature is unit-tested.
- [x] There are end-to-end tests that exercise the feature.

atifaziz:
Closes #9

## Summary of changes:

- Registry now maintains a timestamp per service, which is updated on each announcement.
- Registry has a new method called `prune` to prune the registry by evicting expired entries.
- Registry has a new configuration argument for its constructor (`new`) to configure the TTL.
- If `CHARIOTT_REGISTRY_TTL_SECS` environment variable is defined then it's used to configure the TTL otherwise the default of 15 seconds is used.
- If `CHARIOTT_REGISTRY_TTL_SECS` has a bad value then Chariott panics at start-up.
- If `CHARIOTT_REGISTRY_TTL_SECS` has too low a value (< 2) then it is set to the minimum value of 2 seconds.
- Unit test cases have been added that exercise the eviction on expiration.
- There is an end-to-end test that exercises the eviction on expiration.

~Note that TTL is approximate. The prune loop is not synchronized to the next entry that needs expiring.~


src/registry.rs
```diff
@@ -20,33 +21,62 @@ pub trait RegistryObserver {
...
+    fn default() -> Self {
+        Self { entry_ttl: Duration::from_secs(15) }
+    }
+}
+
 #[derive(Clone, Debug)]
 pub struct Registry<T: RegistryObserver> {
     external_services_by_intent: HashMap<IntentConfiguration, HashSet<ServiceConfiguration>>,
-    known_services: HashSet<ServiceConfiguration>,
+    known_services: HashMap<ServiceConfiguration, SystemTime>,
```
p-schuler, line 48: for my understanding: I didn't see a lookup on the time by configuration. Something we plan to add?

atifaziz: No, there's no look-up. The code does a scan of the map during pruning. `ServiceConfiguration` serves as a key for `HashSet` so one can't put the timestamp in there (besides it not being configuration data). So yes, while the `HashMap` may mislead to think there's some sort of look-up being done, it's really used to pair the variant data together with the invariant part. The other option would be to keep a list _and_ a set of service keys. It would add some overhead without any benefit.

tests/README.md
```diff
@@ -7,12 +7,16 @@ To run the tests, you can either use the CI, or run the tests locally.
...
-2. Run Chariott and the Key-Value app in the background, using
-`./target/release/chariott &` and `./target/release/kv-app &` (assuming you are
-in the repository root). If you want to display debug logs, make sure to set the
-log level via the `RUST_LOG` environment variable.
-3. Execute the tests using `cargo test --test "*e2e"`.
+Assuming the current working directory is the project root, run:
+
+```sh
+cargo build --release -p chariott -p kv-app
+CHARIOTT_REGISTRY_TTL_SECS=2 ./target/release/chariott &
```
bastbu, line 14: How frequently does the key-value application re-announce itself? If it's longer than two seconds, Chariott will evict the key-value application from the registry and it will reregister, causing unreliable E2E tests.

tests/registry-e2e.rs
```diff
@@ -0,0 +1,50 @@
...
+        ExecutionLocality::Local,
+    );
+
+    let mut chariott = setup().await;
+
+    // act
+    builder.register_once(&mut None, true).await?;
+
+    let initial_entries = chariott.inspect("system.registry", namespace.clone()).await?;
+    sleep(Duration::from_secs(5)).await;
```
bastbu, line 38: [nit] Do we want to add a comment around why we wait five seconds? With the context of the PR it's clear to me that we set the TTL to two seconds (though I think we also need to do this for the E2E tests in the CI), but it might be worth mentioning that information.

Cargo.toml
```diff
@@ -33,6 +33,7 @@ chariott-common = { workspace = true }
 prost = { workspace = true }
 prost-types = { workspace = true }
 tokio = { workspace = true, features = ["rt-multi-thread"] }
+tokio-util = { workspace = true }
 tonic = { workspace = true }
 tonic-reflection = "0.5"
 tracing = { workspace = true }
```
bastbu, line 39: Can we make sure to [exclude the new E2E test from the test harness](https://github.com/eclipse/chariott/tree/main/tests#adding-new-tests)? Otherwise it's run when we only want to run unit tests.

src/ext.rs
```diff
@@ -0,0 +1,9 @@
+pub trait OptionExt<T, E> {
```
bastbu, line 1: [nit] We should add the file header to new files.

src/main.rs
```diff
@@ -38,13 +54,55 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    let error_cancellation_token = CancellationToken::new();
+    let ctrl_c_cancellation_token = ctrl_c_cancellation();
+
+    let prune_loop = {
+        use tokio::{select, spawn, time::sleep};
+
+        let ctrl_c_cancellation_token = ctrl_c_cancellation_token.clone();
+        let error_cancellation_token = error_cancellation_token.child_token();
+
+        spawn(async move {
```
bastbu, line 73: Why do we spawn this? We could just return it as a `Future` and then when we call `tokio::join!` it will start to make progress?

src/main.rs
```diff
@@ -38,13 +54,55 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
     let router = router.add_service(reflection_service);
 
-    router.serve_with_ctrl_c_shutdown(addr).await?;
+    let error_cancellation_token = CancellationToken::new();
+    let ctrl_c_cancellation_token = ctrl_c_cancellation();
+
+    let prune_loop = {
+        use tokio::{select, spawn, time::sleep};
+
+        let ctrl_c_cancellation_token = ctrl_c_cancellation_token.clone();
```
bastbu, line 70: [nit] `child_token` would make it more explicit that this is for consuming cancellations and not for generating them, as the tree from the child token cannot affect the parent.

```suggestion
        let ctrl_c_cancellation_token = ctrl_c_cancellation_token.child_token();
```

tests/README.md
atifaziz: Good point, but funny this didn't break locally unless I had my setup wrong. Anyway, Fixed with 28248bc741f87d102e5e76457fa35600d93233d7.

src/ext.rs
```diff
@@ -0,0 +1,9 @@
...
+    fn ok(self) -> Result<Option<T>, E>;
+}
+
+impl<T, E> OptionExt<T, E> for Option<Result<T, E>> {
+    fn ok(self) -> Result<Option<T>, E> {
+        self.map_or(Ok(None), |r| r.map(Some))
+    }
+}
```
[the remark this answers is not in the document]
atifaziz: Fair point. Added with d9ad51448e195ae98a927cf8a0fb7f1bd9814c3e.

Cargo.toml
atifaziz: Done with e483bdbba2abaa572a657a16222d974d48118742.

src/ext.rs
atifaziz: Addressed with 274aae29473ed2a40452621901ddb227d077fb0b.

We should have a CI check for this.

src/main.rs
atifaziz: This was a refactoring oversight. I made `router_serve` a future, but forgot to come back to this one. Thanks for spotting. Addressed with 2bd778b30676bb984efc8d9c151fe3e5b66c85d6.

src/main.rs
bastbu: The side effect of the prune loop accidentally cancelling the `ctrl_c_cancellation_token` would also shut down the server, which is acceptable in this scenario. Consider this comment resolved.

src/main.rs
```diff
@@ -1,11 +1,19 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
+mod ext;
```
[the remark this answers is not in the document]
atifaziz: Addressed with 1dad1154078f1a2ee1e4dcda861b8246ce3db0cf and moved to common with 8b4b05d029e2514342857df5e35279077726ede8.

src/registry.rs
```diff
@@ -20,33 +21,99 @@ pub trait RegistryObserver {
...
 
     /// Returns whether the specified service configuration is already known to
     /// the registry. As system services cannot be updated, invocations with a
     /// system service configuration results in undefined behavior.
     pub fn has_service(&self, key: &ServiceConfiguration) -> bool {
-        self.known_services.contains(key)
+        self.known_services.contains_key(key)
+    }
+
+    pub fn prune(&mut self, timestamp: Instant) -> Option<Instant> {
```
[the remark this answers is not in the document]
atifaziz: Logic reused between `prune` and `upsert` with a1a6bf75e7ce34547b50adc2057e824a6e2ed1e8.

tests/registry-e2e.rs
atifaziz: I made this a bit more obvious and strict with a649b449c2c1e346bee77b2743b4d37661a04efc. Let me know what you think.

src/chariott_grpc.rs
```diff
@@ -35,6 +36,11 @@ impl ChariottServer {
         Self { registry: Arc::new(RwLock::new(registry)), broker }
     }
 
+    pub fn registry_do<T>(&self, f: impl FnOnce(&mut Registry<IntentBroker>) -> T) -> T {
+        let mut registry = self.registry.write().unwrap();
```
p-schuler, line 40: I assume we `blindly` unwrap because we must have had a panic by another thread to run into an Error here?

src/main.rs
```diff
@@ -38,13 +52,54 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+
+    let prune_loop = {
+        let ctrl_c_cancellation_token = ctrl_c_cancellation_token.clone();
+        let error_cancellation_token = error_cancellation_token.child_token();
+
+        async move {
+            tracing::debug!("Prune loop running (TTL = {registry_entry_ttl:?}).");
+            loop {
+                let wakeup_deadline = server.registry_do(|reg| {
+                    reg.prune(Instant::now()).unwrap_or(Instant::now() + registry_entry_ttl)
```
p-schuler, line 73: Why not use unwrap_or_else to lazy evaluate?

src/main.rs
```diff
@@ -38,13 +52,54 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+        Server::builder().add_service(ChariottServiceServer::from_arc(Arc::clone(&server)));
 
     #[cfg(build = "debug")]
     let router = router.add_service(reflection_service);
 
-    router.serve_with_ctrl_c_shutdown(addr).await?;
+    let error_cancellation_token = CancellationToken::new();
+    let ctrl_c_cancellation_token = ctrl_c_cancellation();
+
+    let prune_loop = {
```
p-schuler, line 65: consider moving this logic outside of the main.rs to keep it tidy.

common/src/ext.rs
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Portions Copyright (c) 2014 Jorge Aparicio
+// Licensed under the MIT license.
+
+pub trait OptionExt<T, E> {
+    fn ok(self) -> Result<Option<T>, E>;
+}
+
+impl<T, E> OptionExt<T, E> for Option<Result<T, E>> {
```
p-schuler, line 9: nice addition!

src/registry.rs
```diff
@@ -20,33 +21,93 @@ pub trait RegistryObserver {
...
+}
+
+impl Config {
+    pub const ENTRY_TTL_MIN: Duration = Duration::from_secs(2);
+
+    pub fn entry_ttl(&self) -> Duration {
+        self.entry_ttl
+    }
+
+    pub fn set_entry_ttl_unchecked(self, value: Duration) -> Self {
```
p-schuler, line 36: Why is this called unchecked? It's usually used to indicate that the code does not do validation of the input. For example: https://doc.rust-lang.org/stable/std/primitive.tuple.html#method.get_unchecked

Here we are validating the value against the Max.

bastbu (approved):

src/registry.rs
atifaziz: We are not really validating, which would mean returning a `Result<T, E>` if the value is invalid. We are simply overriding the value from the caller if it's unacceptable and so the caller cannot assume that the method will obey setting the input value exactly as provided. It is _range-bound_. If _unchecked_ doesn't sound right, I could suggest `set_entry_ttl_bounded` as an alternative. I wanted to leave room open for a checked version that's simply called `set_entry_ttl` and returns `Result<Self, InvalidDuration>`.

src/chariott_grpc.rs
atifaziz: Correct. Concerns?

src/registry.rs
p-schuler: Sounds great `set_entry_ttl_bounded` is clearer to me.

src/main.rs
atifaziz: My initial thinking was because it's not expensive to compute and I wanted to avoid a closure with `Instant::now()`, but digging into this now, I see that everything gets in-lined without any closure. Here's a simple test I tried:

```rust
use std::time::*;

pub fn test1() -> Instant {
    let now = Instant::now();
    prune(now).unwrap_or(now + Duration::from_secs(5))
}

pub fn test2() -> Instant {
    let now = Instant::now();
    prune(now).unwrap_or_else(|| now + Duration::from_secs(5))
}

#[inline(never)]
pub fn prune(time: Instant) -> Option<Instant> {
    None
}
```

which gets compiled to the following:

```asm
example::test1:
        push    r14
        push    rbx
        sub     rsp, 24
        call    qword ptr [rip + std::time::Instant::now@GOTPCREL]
        mov     r14, rax
        mov     rbx, rdx
        mov     rdi, rsp
        call    qword ptr [rip + example::prune@GOTPCREL]
        mov     edx, 5
        mov     rdi, r14
        mov     rsi, rbx
        xor     ecx, ecx
        call    qword ptr [rip + <std::time::Instant as core::ops::arith::Add<core::time::Duration>>::add@GOTPCREL]
        cmp     qword ptr [rsp], 0
        je      .LBB0_2
        mov     rdx, qword ptr [rsp + 16]
        mov     rax, qword ptr [rsp + 8]
.LBB0_2:
        add     rsp, 24
        pop     rbx
        pop     r14
        ret

example::test2:
        push    r14
        push    rbx
        sub     rsp, 24
        call    qword ptr [rip + std::time::Instant::now@GOTPCREL]
        mov     r14, rax
        mov     rbx, rdx
        mov     rdi, rsp
        call    qword ptr [rip + example::prune@GOTPCREL]
        cmp     qword ptr [rsp], 0
        je      .LBB1_1
        mov     rax, qword ptr [rsp + 8]
        mov     rdx, qword ptr [rsp + 16]
        jmp     .LBB1_3
.LBB1_1:
        mov     edx, 5
        mov     rdi, r14
        mov     rsi, rbx
        xor     ecx, ecx
        call    qword ptr [rip + <std::time::Instant as core::ops::arith::Add<core::time::Duration>>::add@GOTPCREL]
.LBB1_3:
        add     rsp, 24
        pop     rbx
        pop     r14
        ret
```

One can see in `test2` that after the call to `prune`:

```asm
        call    qword ptr [rip + example::prune@GOTPCREL]
```

it checks for `None`:

```asm
        cmp     qword ptr [rsp], 0
```

and [if it's the case, then simply _jumps_](https://en.wikibooks.org/wiki/X86_Assembly/Control_Flow#Jump_if_Equal) to a label:

```asm
        je      .LBB1_1
```

The code at that label then does the `now + Duration::from_secs(5)` bit:

```asm
.LBB1_1:
        mov     edx, 5
        mov     rdi, r14
        mov     rsi, rbx
        xor     ecx, ecx
        call    qword ptr [rip + <std::time::Instant as core::ops::arith::Add<core::time::Duration>>::add@GOTPCREL]
```

I've therefore gone ahead and used `unwrap_or_else` instead with 09607fda01080ac3b8d3d1094ecd0316f94654eb.

src/chariott_grpc.rs
p-schuler: No - just confirming.

src/registry.rs
p-schuler: Makes sense. It was a bit confusing as I thought we do ttl lookup by ServiceConfiguration.

src/main.rs
atifaziz: Done with a1b5b40955432bfd2cf05d25a3af09413098b91d.

PR #15, opened 2022-11-01
Support subscription to registry changes

### LINKED ISSUE #8: Allow applications to subscribe to registry changes
Applications should be able to subscribe to the registry for notification of changes, such as when applications register intents with Chariott.

An application can subscribe for keys formatted as `system.registry` `namespaces/NAMESPACE` where `NAMESPACE` is the name of some namespace, e.g. `system.registry` `namespaces/sdv.kvs`. Whenever something related to the namespace changes, Chriott should generate an event that is a string indicating the change (like added or removed). It is legal for an application to subscribe to changes in a namespace even if it's not currently registered.

This depends on issue #9.

## Acceptance Criteria

- [x] The `system.registry` namespace supports the subscribe intent.
- [x] An application can subscribe to changes for a namespace in the Chariott registry.
- [x] A subscribed application does not receive a notification if just end-point changes take place.
- [x] The feature is unit-tested.
- [x] There are end-to-end tests that exercise the feature.

bastbu:
Closes #8 

## Motivation and Context

As demonstrated in the dog mode logic application, applications can use the `Inspect` intent on the registry to adapt their behaviour based on the presence or absence of certain intents for a namespace.

This is more useful if an application gets notifications from the registry that the available intents for a certain namespace changed. If the application receives a notification that one of the relevant namespaces has a different set of intents, the application can re-inspect the registry and adapt its behaviour based on the updated content of the registry. 

## Description

To support the feature above, the registry passes more detailed information to the observer in form of a sequence of `Change<'a>`. By introducing another observer, which connects the observed changes to the event sub-system (ESS), subscribers are notified on relevant changes to the registry.

Relevant changes encompass changes that either _add_ or _remove_ intents for a namespace, but not ones that only _modify_ service registrations for an already provided intent.

As per acceptance criteria:

- The applications can subscribe to `system.registry` with keys of the form `namespaces/<namespace>` to be notified of changes to a namespace.
- The subscription is allowed whether or not the namespace is currently registered in the registry

Different from the acceptance criteria:
- The event payload on registry changes is empty. We could list all intents that changed for a namespace together with whether they were added or removed, but the applications are likely going to re-inspect the registry and as such this could be wasted additional complexity.


examples/common/src/chariott/proto.rs
```diff
@@ -1,14 +1,7 @@
...
-    // and https://github.com/tokio-rs/prost/issues/661#issuecomment-1156606409
-    // why we use allow derive_partial_eq_without_eq
-    #![allow(clippy::derive_partial_eq_without_eq)]
-    tonic::include_proto!("chariott.streaming.v1");
-}
-
 pub use chariott_common::proto::common;
 pub use chariott_common::proto::provider;
 pub use chariott_common::proto::runtime as runtime_api;
+pub use chariott_common::proto::streaming;
```
bastbu, line 7: For the reviewer: this file can now be removed entirely. I will do this as a follow-up PR to not generate unnecessary noise on this PR.

src/main.rs
```diff
@@ -26,20 +32,31 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     collector.init();
 
-    let broker = IntentBroker::new();
-    let registry = Registry::new(broker.clone());
+    let ess = SharedEss::new();
+    let broker = IntentBroker::new(
+        format!(
+            "http://{}:{}",
+            env(EXTERNAL_HOST_NAME_ENV).unwrap_or_else(|| "localhost".to_string()),
```
atifaziz, line 39: You can avoid allocating the `"localhost"` string given that it's a literal:

```suggestion
            env::<String>(EXTERNAL_HOST_NAME_ENV).as_deref().unwrap_or("localhost"),
```

common/src/ess.rs
```diff
@@ -0,0 +1,188 @@
...
+
+        let id = Uuid::new_v4().to_string();
+        let (_, receiver_stream) = self.0.read_events(id.clone().into());
+        let mut response = Response::new(receiver_stream);
+        response.metadata_mut().insert(METADATA_KEY, id.try_into().unwrap());
+        Ok(response)
+    }
+}
+
+impl<T> AsRef<InnerEss<T>> for SharedEss<T> {
```
atifaziz, line 91: I think what you want to implement here is `Deref` instead. Then you don't have to see all those `.as_ref()` calls on the call sites (in other words, from the caller's perspective, the wrapper appears to be like its inner type):

```diff
diff --git a/common/src/ess.rs b/common/src/ess.rs
index 662aa47..0c470a7 100644
--- a/common/src/ess.rs
+++ b/common/src/ess.rs
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
-use std::{sync::Arc, time::SystemTime};
+use std::{ops::Deref, sync::Arc, time::SystemTime};
 
 use crate::proto::{
     common::Value as ValueMessage,
@@ -88,8 +88,10 @@ where
     }
 }
 
-impl<T> AsRef<InnerEss<T>> for SharedEss<T> {
-    fn as_ref(&self) -> &InnerEss<T> {
+impl<T> Deref for SharedEss<T> {
+    type Target = InnerEss<T>;
+
+    fn deref(&self) -> &Self::Target {
         self.0.as_ref()
     }
 }
diff --git a/examples/common/src/chariott/streaming.rs b/examples/common/src/chariott/streaming.rs
index 99ae11a..79e4250 100644
--- a/examples/common/src/chariott/streaming.rs
+++ b/examples/common/src/chariott/streaming.rs
@@ -18,7 +18,7 @@ struct Ess<T>(InnerEss<(EventId, T)>);
 
 impl<T: Clone + Send + 'static> Observer<EventId, T> for Ess<T> {
     fn on_set(&mut self, key: &EventId, value: &T) {
-        self.0.as_ref().publish(key, (key.clone(), value.clone()));
+        self.0.publish(key, (key.clone(), value.clone()));
     }
 }
 
diff --git a/src/registry.rs b/src/registry.rs
index 54b75bf..9917f4f 100644
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -36,7 +36,7 @@ impl Observer for Ess {
             })
             .collect::<HashSet<_>>()
         {
-            self.as_ref().publish(format!("namespaces[{}]", namespace).as_str(), ());
+            self.publish(format!("namespaces[{}]", namespace).as_str(), ());
         }
     }
 }
```

examples/common/src/chariott/streaming.rs
```diff
@@ -1,49 +1,30 @@
...
+impl<T: Clone + Send + 'static> Observer<EventId, T> for Ess<T> {
+    fn on_set(&mut self, key: &EventId, value: &T) {
+        self.0.as_ref().publish(key, (key.clone(), value.clone()));
     }
 }
 
-impl<T: Clone> Observer<EventId, T> for Arc<Ess<T>> {
-    fn on_set(&mut self, key: &EventId, value: &T) {
-        self.publish(key, (key.clone(), value.clone()));
+impl<T> AsRef<InnerEss<(EventId, T)>> for Ess<T> {
```
atifaziz, line 25: Like the [other suggestion](https://github.com/eclipse/chariott/pull/15/files#r1012780174), I think you want to implement `Deref` here too:

```diff
diff --git a/examples/common/src/chariott/streaming.rs b/examples/common/src/chariott/streaming.rs
index 99ae11a..848ce52 100644
--- a/examples/common/src/chariott/streaming.rs
+++ b/examples/common/src/chariott/streaming.rs
@@ -6,7 +6,7 @@ use chariott_common::{
     proto::common::{fulfillment::Fulfillment, ReadFulfillment, ReadIntent, SubscribeIntent},
 };
 use keyvalue::{InMemoryKeyValueStore, Observer};
-use std::sync::RwLock;
+use std::{ops::Deref, sync::RwLock};
 use tonic::Status;
 
 use crate::chariott::proto::{common::value::Value as ProtoValue, common::Value as ValueMessage};
@@ -22,8 +22,10 @@ impl<T: Clone + Send + 'static> Observer<EventId, T> for Ess<T> {
     }
 }
 
-impl<T> AsRef<InnerEss<(EventId, T)>> for Ess<T> {
-    fn as_ref(&self) -> &InnerEss<(EventId, T)> {
+impl<T> Deref for Ess<T> {
+    type Target = InnerEss<(EventId, T)>;
+
+    fn deref(&self) -> &Self::Target {
         &self.0
     }
 }
@@ -82,7 +84,7 @@ where
     T: Into<ProtoValue> + Clone + Send + Sync + 'static,
 {
     fn subscribe(&self, subscribe_intent: SubscribeIntent) -> Result<Fulfillment, Status> {
-        let result = self.ess.as_ref().serve_subscriptions(subscribe_intent, |(_, v)| v.into())?;
+        let result = self.ess.serve_subscriptions(subscribe_intent, |(_, v)| v.into())?;
         Ok(Fulfillment::Subscribe(result))
     } 
```

examples/common/src/chariott/streaming.rs
```diff
@@ -111,22 +66,9 @@ where
...
-    async fn open(
-        &self,
-        _: tonic::Request<OpenRequest>,
-    ) -> Result<Response<Self::OpenStream>, Status> {
-        let id: ClientId = uuid::Uuid::new_v4().to_string().into();
-        let (_, receiver_stream) = self.ess.read_events(id.clone());
-        let mut response = Response::new(receiver_stream);
-        response.metadata_mut().insert(METADATA_KEY, id.to_string().try_into().unwrap());
-        Ok(response)
+impl<T> AsRef<InnerEss<(EventId, T)>> for StreamingStore<T> {
```
atifaziz, line 69: I don't think a `StreamingStore<T>` should _pretend_ to be an ESS. I would just provide a simple getter that's clear:

```rust
impl<T> StreamingStore<T> {
    pub fn ess(&self) -> &InnerEss<(EventId, T)> {
        &self.ess.0
    }
}
```

bastbu: Thanks for the suggestion, done with 1ce2e6ef20a5b9e76f41fc82675bcf9fcab5423a.

examples/common/src/chariott/streaming.rs
bastbu: Based on your other comment, I removed the `AsRef` instead with 1ce2e6ef20a5b9e76f41fc82675bcf9fcab5423a as the newtype becomes a private implementation detail.

common/src/ess.rs
bastbu: Reconsidering this: that the `SharedEss` _is_ not a smart pointer but _contains_ a smart pointer is a private implementation detail. Seen from the outside it still behaves as if it were a smart pointer, hence I applied your suggestion with 11c04e890f14980efa2adb5137c2b364624ff681.

common/src/ess.rs
```diff
@@ -0,0 +1,190 @@
...
+        subject
+            .serve_subscriptions(
+                SubscribeIntent { channel_id, sources: vec![EVENT_A.into(), EVENT_B.into()] },
+                |_| ValueEnum::Null(0),
+            )
+            .unwrap();
+
+        // assert
+        subject.0.publish(EVENT_A, ());
+        subject.0.publish(EVENT_B, ());
```
atifaziz, line 146: nit:

```suggestion
        subject.publish(EVENT_A, ());
        subject.publish(EVENT_B, ());
```

atifaziz: Don't these belong in the act part?

common/src/ess.rs
```diff
@@ -0,0 +1,190 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+use std::{sync::Arc, time::SystemTime, ops::Deref};
```
atifaziz, line 4: `rust-fmt` insists: 😄

```suggestion
use std::{ops::Deref, sync::Arc, time::SystemTime};
```

common/src/ess.rs
bastbu: I don't think so, as these are events only published for assertions. That was the only way I saw to check whether we actually served the correct subscription. Applied your suggesion.

common/src/ess.rs
```diff
@@ -0,0 +1,190 @@
...
+    streaming::{channel_service_server::ChannelService, Event, OpenRequest},
+};
+use async_trait::async_trait;
+use ess::EventSubSystem;
+use tokio::spawn;
+use tokio_stream::wrappers::ReceiverStream;
+use tonic::{Response, Status};
+use uuid::Uuid;
+
+type InnerEss<T> = EventSubSystem<Box<str>, Box<str>, T, Result<Event, Status>>;
```
atifaziz, line 18: Can we rename `InnerEss` to `Ess` or (preferably) `EventSubSystem`? I don't think the _inner_ labeling is right. This is just specifying some type arguments and leaving one open to parameterisation.

common/src/ess.rs
```diff
@@ -0,0 +1,190 @@
...
+}
+
+impl<T: Clone + Send + 'static> SharedEss<T> {
+    pub fn serve_subscriptions(
+        &self,
+        subscribe_intent: SubscribeIntent,
+        into_value: fn(T) -> ValueEnum,
+    ) -> Result<SubscribeFulfillment, Status> {
+        let subscriptions = self
+            .0
```
atifaziz, line 46: nit: this can be removed:

```suggestion
```

p-schuler:
General comment on the description:

>The event payload on registry changes is empty. We could list all intents that changed for a namespace together with whether they were added or removed, but the applications are likely going to re-inspect the registry and as such this could be wasted additional complexity.

I agree not listing the intents makes sense. I assume the source namespace is still communicated (as we register for multiple keys, only one namespace is triggering the change).

bastbu:
> General comment on the description:
> 
> > The event payload on registry changes is empty. We could list all intents that changed for a namespace together with whether they were added or removed, but the applications are likely going to re-inspect the registry and as such this could be wasted additional complexity.
> 
> I agree not listing the intents makes sense. I assume the source namespace is still communicated (as we register for multiple keys, only one namespace is triggering the change).

Thanks for your input. This is indeed the case: the namespace is delivered via the [source key](https://github.com/eclipse/chariott/blob/main/proto/chariott/streaming/v1/streaming.proto#L45) of the event payload.

p-schuler:
> > General comment on the description:
> > > The event payload on registry changes is empty. We could list all intents that changed for a namespace together with whether they were added or removed, but the applications are likely going to re-inspect the registry and as such this could be wasted additional complexity.
> > 
> > 
> > I agree not listing the intents makes sense. I assume the source namespace is still communicated (as we register for multiple keys, only one namespace is triggering the change).
> 
> Thanks for your input. This is indeed the case: the namespace is delivered via the [source key](https://github.com/eclipse/chariott/blob/main/proto/chariott/streaming/v1/streaming.proto#L45) of the event payload, and will be something like `namespaces[sdv.vdt]`.

We should re-consider the formatting here. I understand that we want to have a way to express the subscription domain (namespaces) and inside the domain the target. Generally, we should try to come up with a concept to handle multiple target domains for subscriptions with a stronger typed fashion. We could add a domain to the subscription interface indicating a sub area for the subscription to target. It would be an optional indicator and the applications would implement a default if it were left out. The event then would indicate the area the event was fired on.

common/src/ess.rs
bastbu: Done with d7deb59057a7f2d53035d6f13df0bed7775f0223.

common/src/ess.rs
```diff
@@ -0,0 +1,189 @@
...
+            vec![EVENT_A, EVENT_B],
+            result.iter().map(|e| e.source.clone()).collect::<Vec<_>>()
+        );
+
+        // assert sequence numbers
+        assert_eq!(1, result[0].seq);
+        assert_eq!(1, result[1].seq);
+
+        // assert payload
+        assert_eq!(Some(ValueMessage { value: Some(ValueEnum::Null(0)) }), result[0].value);
```
atifaziz: nit: Instead of different assert blocks, this could just be a loop over expected sources and other data being constant?

```suggestion
        let expected_sources = [EVENT_A, EVENT_B];

        assert_eq!(expected_sources.len(), result.len());

        for (expected_source, actual) in expected_sources.iter().zip(result) {
            assert_eq!(*expected_source, actual.source.as_str());
            assert_eq!(1, actual.seq);
            assert_eq!(Some(ValueMessage { value: Some(ValueEnum::Null(0)) }), actual.value);
        }
```

tests/registry-e2e.rs
```diff
@@ -46,6 +47,40 @@ async fn expired_registrations_are_pruned_after_ttl() -> Result<(), anyhow::Erro
...
+        format!("namespaces/{}", namespace)
+    }
+
+    // arrange
+    let namespace = format!("e2e.registration.{}", get_uuid());
+
+    let builder = RegistrationBuilder::new(
+        "registration.provider.e2e",
+        "1.0.0",
+        "http://test-url:7090".parse().unwrap(), // arbitrary url, the provider will never be invoked
```
atifaziz, line 62: nit: Why not `localhost` instead of `test-url`?

common/src/ess.rs
bastbu: Thanks, applied your suggestion with a slight modification in bc8dbc96723ac6faa5869a57f355b95e881e8a66.

tests/registry-e2e.rs
bastbu: Fixed this with a20a5330cf913881675d6eb0146c85d9e62a24b8.

tests/registry-e2e.rs
```diff
@@ -46,6 +47,40 @@ async fn expired_registrations_are_pruned_after_ttl() -> Result<(), anyhow::Erro
...
+    // act
+    let stream =
+        subject.listen("system.registry", vec![namespace_event(&namespace).into()]).await?;
+
+    builder.register_once(&mut None, true).await?;
+
+    // assert
+    let stream = stream.timeout(Duration::from_secs(5)).take(1).collect::<Vec<_>>().await;
+    let result = stream.into_iter().next().unwrap();
+    assert_eq!(namespace_event(&namespace).as_str(), result.unwrap().unwrap().id.as_ref());
```
atifaziz, line 79: nit: you can remove all the `as_ref()`s and `into()` in the test code if you change the return type on the local helper:

```suggestion
    fn namespace_event(namespace: &str) -> Box<str> {
        format!("namespaces/{}", namespace).into()
    }

    // arrange
    let namespace = format!("e2e.registration.{}", get_uuid());

    let builder = RegistrationBuilder::new(
        "registration.provider.e2e",
        "1.0.0",
        "http://localhost:7090".parse().unwrap(), // arbitrary url, the provider will never be invoked
        &namespace,
        [Intent::Inspect],
        ExecutionLocality::Local,
    );

    let mut subject = setup().await;

    // act
    let stream = subject.listen("system.registry", vec![namespace_event(&namespace)]).await?;

    builder.register_once(&mut None, true).await?;

    // assert
    let stream = stream.timeout(Duration::from_secs(5)).take(1).collect::<Vec<_>>().await;
    let result = stream.into_iter().next().unwrap();
    assert_eq!(namespace_event(&namespace), result.unwrap().unwrap().id);
```

atifaziz (approved):

PR #32, opened 2022-11-10
Homogenize imports from proto-generated modules

bastbu:
## Motivation and Context

As a follow-up from #15, we can remove an unnecessary re-export of `proto.rs` submodules from `examples_common` in favor of importing them directly from `chariott_common`.

## Description

We use this opportunity to homogenize the imports from the protoc-generated Rust modules. While before the PR we use a mix of qualifiers and imports, we now consistently import the protoc-generated types, which implies fewer qualifiers and noise within the code.

For `Fulfillment` and `Intent` protoc generates a nested Rust structure generated for said types. We consistently name the outer fulfillment `FulfillmentMessage`, and the inner fulfillment `FulfillmentEnum` where a name conflict exists.


atifaziz (commented):
Certainly much cleaner, but I'd like to suggest going one step further and extracting `proto.rs` into a stand-alone crate called `chariott-proto` (sitting in the `proto` directory we already have), that's separate from `chariott-common`. The former would be the crate representing the official Rust stubs for the Chariott APIs and the latter just a crate for common code. This way, `chariott-proto` can be used separate from `chariott-common`. It would also avoid _common_ appearing twice in imports, as in `chariott_common::proto::common::*`, which reads verbose and confusing. Once we have `chariott-proto`, we can also do additional things:

- Directly add aliases like `FilfillmentMessage` and `FulfillmentEnum` so that they don't have to be repeated everywhere.
- Expose a more logical tree of symbols so we're independent of the `*.proto` file structure. For example, the types in `common.proto` could be directly exposed under `chariott_proto`.

atifaziz (commented):
I would recommend reverting the `common` and `proto` folder renames. Is the `chariott-` prefix adding much value/clarity? I imagine reverting the renames would reduce the number of changes in the commit.

chariott-proto/src/lib.rs
```diff
@@ -10,6 +10,12 @@ mod chariott {
     pub mod common {
         pub mod v1 {
             tonic::include_proto!("chariott.common.v1");
+            pub use fulfillment::Fulfillment as FulfillmentEnum;
```
atifaziz, line 13: Didn't we say, these will be exposed directly under `chariott_proto` as opposed to `chariott_proto::common`?

bastbu:
> I would recommend reverting the `common` and `proto` folder renames. Is the `chariott-` prefix adding much value/clarity? I imagine reverting the renames would reduce the number of changes in the commit.

Reverted said commit and renamed `chariott-proto` to `proto.rs` instead to avoid the directory conflicts in ecf473008de709407ec5b6f50bffc5fa00e286c6

examples/applications/mock-vas/src/chariott_provider.rs
```diff
@@ -6,6 +6,13 @@ use std::sync::Arc;
 use std::vec;
 
 use async_trait::async_trait;
+use chariott_proto::{
+    common::{
+        discover_fulfillment::Service, DiscoverFulfillment, FulfillmentEnum, FulfillmentMessage,
+        IntentEnum, InvokeFulfillment, Value as ProtoValue,
```
atifaziz, line 12: Do we not want to import and use `ValueMessage` here instead of aliasing `Value` as `ProtoValue`?

atifaziz: Perhaps some more cases to review?

    $ git grep -E '\bas\s+Proto'
    examples/common/src/chariott/api.rs:        discover_fulfillment::Service as ProtoService, DiscoverFulfillment, DiscoverIntent,
    examples/common/src/chariott/inspection.rs:    fulfillment::Fulfillment, inspect_fulfillment::Entry as ProtoEntry, InspectFulfillment,
    examples/common/src/examples/detection.rs:    DetectEntry, DetectRequest as ProtoDetectRequest, DetectResponse as ProtoDetectResponse,

examples/common/src/chariott/api.rs
```diff
@@ -11,18 +11,19 @@ use std::{
...
-    proto::{
-        common::{
-            fulfillment::Fulfillment as ProtoFulfillment, DiscoverFulfillment, InspectFulfillment,
-            InvokeFulfillment, ReadFulfillment, SubscribeFulfillment, WriteFulfillment,
-        },
-        runtime::FulfillResponse,
+use chariott_common::error::{Error, ResultExt as _};
+use chariott_proto::{
+    common::{
+        discover_fulfillment::Service as ProtoService, DiscoverFulfillment, DiscoverIntent,
```
atifaziz, line 20: nit: So many types here so may be import everything via `*`? And then perhaps the `ProtoService` exception can be move closer to where it's used/needed? For example:

```diff
diff --git a/examples/common/src/chariott/api.rs b/examples/common/src/chariott/api.rs
index 78c0a698..754f88a7 100644
--- a/examples/common/src/chariott/api.rs
+++ b/examples/common/src/chariott/api.rs
@@ -384,8 +384,10 @@ pub struct Service {
     pub schema_reference: Box<str>,
 }
 
-impl From<ProtoService> for Service {
-    fn from(value: ProtoService) -> Self {
+use chariott_proto::common::discover_fulfillment::Service as DiscoverFulfillmentService;
+
+impl From<DiscoverFulfillmentService> for Service {
+    fn from(value: DiscoverFulfillmentService) -> Self {
         Service {
             url: value.url.into(),
             schema_kind: value.schema_kind.into(),
```

chariott-proto/src/lib.rs
bastbu: We could, but we would still need to re-export the original common module as it's used by https://github.com/eclipse/chariott/blob/main/examples/common/src/examples/proto.rs. There might be another way to solve this (recompiling the common proto), but not sure it's worth the effort at this point.

examples/common/src/chariott/inspection.rs
```diff
@@ -3,11 +3,9 @@
...
 use std::{borrow::Borrow, collections::HashMap};
 
-use chariott_common::{
-    proto::common::{
-        fulfillment::Fulfillment, inspect_fulfillment::Entry as ProtoEntry, InspectFulfillment,
-    },
-    query::regex_from_query,
+use chariott_common::query::regex_from_query;
+use chariott_proto::common::{
+    fulfillment::Fulfillment, inspect_fulfillment::Entry as ProtoEntry, InspectFulfillment,
```
atifaziz, line 8: nit: Should we rename `ProtoEntry` to `EntryMessage` for consistency?

examples/common/src/chariott/registration.rs
```diff
@@ -3,20 +3,20 @@
...
-
-use crate::chariott::proto::runtime_api::{
-    chariott_service_client::ChariottServiceClient, intent_service_registration::ExecutionLocality,
-    *,
+use chariott_proto::runtime::{
+    chariott_service_client::ChariottServiceClient, intent_registration::Intent,
+    intent_service_registration::ExecutionLocality, AnnounceRequest, IntentRegistration,
+    IntentServiceRegistration, RegisterRequest, RegistrationState,
 };
+use tokio::time::sleep;
```
atifaziz, line 15: nit: Intentional move?

examples/applications/simulated-camera/src/communication.rs
```diff
@@ -4,19 +4,18 @@
...
 use chariott_common::{
     error::{Error, ResultExt as _},
     shutdown::{ctrl_c_cancellation, RouterExt as _},
 };
-use examples_common::chariott::proto::{
+use chariott_proto::{
     provider::provider_service_server::ProviderServiceServer,
     streaming::channel_service_server::ChannelServiceServer,
 };
+use futures::future::join_all;
```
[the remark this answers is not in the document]
bastbu: Intentional, yes, but not necessarily required.

Since we used the convention to have all usings (except `crate`, `super` and `std`) in a single block most frequently, I did the same here. I was already adding a new `use` statement (`use chariott_proto::`), hence I would have needed to decide in which of the two blocks to put it. Instead, I merged the two blocks, which made it clear where to put the new `use` statement.

examples/common/src/examples/detection.rs
```diff
@@ -2,11 +2,14 @@
...
 use chariott_common::error::{Error, ResultExt};
+use chariott_proto::common::{Blob, InvokeFulfillment, InvokeIntent};
 use prost::Message;
 
-use crate::chariott::{proto::common as common_proto, value::Value};
+use crate::chariott::value::Value;
 
-use super::proto::detection as detection_proto;
+use super::proto::detection::{
+    DetectEntry, DetectRequest as ProtoDetectRequest, DetectResponse as ProtoDetectResponse,
```
atifaziz, line 11: nit: Consider renaming the aliases from `Proto*` to `*Message` for consistency.

proto.rs/src/lib.rs
```diff
@@ -10,6 +10,12 @@ mod chariott {
     pub mod common {
         pub mod v1 {
             tonic::include_proto!("chariott.common.v1");
+            pub use fulfillment::Fulfillment as FulfillmentEnum;
```
atifaziz, line 13: nit: Worth adding a comment why this is done so?

src/chariott_grpc.rs
```diff
@@ -4,6 +4,14 @@
 use std::sync::{Arc, RwLock};
 use std::time::Instant;
 
+use chariott_proto::{
+    common::intent::Intent,
+    runtime::{
+        chariott_service_server::ChariottService, AnnounceRequest, AnnounceResponse,
+        FulfillRequest, FulfillResponse, IntentRegistration, IntentServiceRegistration,
+        RegisterRequest, RegisterResponse, RegistrationState,
+    },
```
atifaziz, line 13: nit: Since it imports so many types directly linked to a module to do with gRPC, perhaps worth simplifying to:

```suggestion
    runtime::{chariott_service_server::ChariottService, *},
```

atifaziz (approved):
Great, clean-up! Thanks!

examples/common/src/chariott/registration.rs
bastbu: Yes, I applied the same reasoning as in https://github.com/eclipse/chariott/pull/32#discussion_r1022562625.

I would suggest to keep it that way, as it makes it clear where to add new `use` statements.

proto.rs/src/lib.rs
bastbu: Done in f56af3343b09e05ccf13fe9ae3e483ece5f4e851

examples/common/src/examples/detection.rs
bastbu: Done in 44ce11f2310210263841d7709e36adec95269785

examples/common/src/chariott/inspection.rs
bastbu: Done in 44ce11f2310210263841d7709e36adec95269785.

examples/applications/mock-vas/src/chariott_provider.rs
bastbu: Done in 44ce11f2310210263841d7709e36adec95269785.

examples/applications/simulated-camera/src/communication.rs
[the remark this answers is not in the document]
atifaziz: I don't think we did this consistently, but whenever possible, I've tried to go with the grouping (separate blocks): imports from `std`, imports from crates and then imports from `crate`. I believe `rustfmt` preserves groups.

We can address the import order consistency in another PR at another time. Meanwhile, I was just wondering if this unrelated change was for a reason.

examples/common/src/chariott/api.rs
bastbu: I removed the `ProtoService` exception in favor of `ServiceMessage`. I tend not to use glob imports, for reasons as described [in the Clippy lints](https://rust-lang.github.io/rust-clippy/master/index.html#wildcard_imports) and [this thread](https://mail.mozilla.org/pipermail/rust-dev/2014-March/thread.html#8981) (the argument from [Bob Ippolito](https://mail.mozilla.org/pipermail/rust-dev/2014-March/008999.html) makes sense to me).

src/chariott_grpc.rs
bastbu: Thanks for the suggestion, I would [prefer to not use glob imports](https://github.com/eclipse/chariott/pull/32#discussion_r1022596092)

PR #37, opened 2022-11-14
Upgrade to Rust 1.65

bastbu:
## Description

Updates the used Rust version to 1.65.


examples/applications/lt-consumer/src/main.rs
```diff
@@ -225,7 +225,7 @@ async fn evaluate_docker_stats(
                 let cpu_delta =
                     cpu_stats.cpu_usage.total_usage - stats.precpu_stats.cpu_usage.total_usage;
 
-                if number_cpus == None {
+                if number_cpus.is_none() {
```
atifaziz, line 228: I'm guessing Clippy complained about this?

examples/applications/simulated-camera/src/camera.rs
```diff
@@ -123,9 +123,9 @@ impl CameraLogic {
     }
 }
 
-// In 1.64 clippy checks get_first
```
atifaziz, line 126: Was removal of this intentional? Does it still apply?

atifaziz (approved):

examples/applications/lt-consumer/src/main.rs
bastbu: Exactly, though I did not investigate why it did not complain before.

examples/applications/simulated-camera/src/camera.rs
bastbu: Yes, it still applies and was also intentional. The updated comment:

```
// try_stream macro is still using accessing first element with
// `$crate::async_stream_impl::try_stream_inner!(($crate) $($tt)*).get(0)`
// https://rust-lang.github.io/rust-clippy/master/index.html#get_first
```

contains the same information as before, but no longer mentions with which version this was introduced. Otherwise, I would have needed to explicitly mention `1.64` and `1.65` as it applies to both versions, instead I future-proofed it by removing any mention of a version.

atifaziz: > Otherwise, I would have needed to explicitly mention `1.64` and `1.65` as it applies to both versions, instead I future-proofed it by removing any mention of a version.

Understandable, which is why I was expecting something like _1.64 or later_ or just _1.64+_.

bastbu: Valid point, I pushed d7d9b619e16dc83185b6d91b52357b637eea61e6 to have a less noisy change

p-schuler (approved):
Looks good.

PR #38, opened 2022-11-14
Do not install Tarpaulin from Devcontainer

### LINKED ISSUE #33: Dev container does not work on ARM based machine
When starting the devcontainer on a MacBook Pro M1 machine, the devcontainer fails. The problem is that tarpaulin does not work with ARM based machines. From the tarpaulin repo:

```text
On Linux Tarpaulin's default tracing backend is still Ptrace and will only work on x86_64 processors. This can be changed to the llvm coverage instrumentation with --engine llvm, for Mac and Windows this is the default collection method.
```

Can we remove the last line in the Dockerfile and add documentation that if you wanted to test code coverage, you need to install that tool due to platform difference?

## Expected Behavior
devcontainer should start on ARM machine

## Current Behavior
devcontainer fails to start because it will not build

## Steps to Reproduce

1. Get an ARM based laptop
2. Start devcontainer

cc: @juliays

bastbu:
Fixes #33

## Description

Tarpaulin [cannot be compiled on Docker on M1 MacBooks](https://github.com/xd009642/tarpaulin/issues/549#issuecomment-1198600914) but has an [(unreleased) fix ready for this issue](https://github.com/xd009642/tarpaulin/issues/549#issuecomment-1291096269). As long as it's not yet released, we do not install Tarpaulin with the Devcontainer.

We track reinstalling Tarpaulin in #39.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -42,7 +42,3 @@ COPY ../rust-toolchain .
 RUN rustup override set $(cat rust-toolchain) && \
     rustup component add clippy && \
     rustup component add rustfmt
-
-# The following takes a long time to complete.
-# We use a separate layer to hit the (layer) cache if possible.
-RUN cargo install cargo-tarpaulin


atifaziz (commented):
Anyway we can make its install conditional instead of completely removing it?

atifaziz (approved):

PR #52, opened 2022-11-17
Request/response fulfillment for cloud applications

### LINKED ISSUE #51: Car fulfills a non-subscription intent for the cloud
## Description

The Car Bridge should be able to fulfill `FulfillRequest`s coming from the cloud by running them against Chariott and publishing the responses as messages to the MQTT broker.

## Acceptance criteria 

- [x] Car bridge proxies (non-subscription) fulfill requests from the cloud to Chariott
- [x] Car bridge uses Request/Response MQTT feature to send the response to MQTT broker

bastbu:
Resolves #51

## Motivation and Context

The Car Bridge must be able to fulfill requests in the car on-behalf of cloud applications.

## Description

To achieve this, the Car Bridge listens to incoming cloud-to-device messages (published on a topic called `c2d/<vin-number>`). The Car Bridge assumes that the payload of these messages is always a `FulfillRequest`. After deserializing the messages, it fulfills all non-subscription and non-discover requests against Chariott and publishes the response to the MQTT broker.

With this PR we also introduce a simple error model. Errors are sent as responses to the MQTT broker, with a user-defined property `"error"` set to `"1"` and a content type which reflects that it is a `Value::String`.

Several issues left open with this PR:

- How do we ensure that if sessions are expired the client resubscribes to the correct topics?
- Testing


car-bridge/src/main.rs
```diff
@@ -29,58 +36,123 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+            .properties()
+            .get_binary(PropertyCode::CorrelationData)
+            .ok_or_else(|| Error::new("No correlation data found on message."))?;
+
+        let response_topic = message
+            .properties()
+            .get_string(PropertyCode::ResponseTopic)
+            .ok_or_else(|| Error::new("No correlation data found on message."))?;
+
+        let fulfill_request: FulfillRequest = Message::decode(message.payload())?;
```
atifaziz, line 122: From this point on, all uses of `?` will cause the function to return on error and the request will have no response generated. Since we have the correlation and response topic, we should be able to generate an error response. Otherwise, the cloud app on the other will only be able to timeout.

bastbu: Agreed, which is why I left it as an open point with this PR:

> How do we report errors back to the cloud?

Do you think we should address it right away, or could this be something for a follow-up PR (together with coming up with a strategy _how_ the error payloads should look like?

bastbu: Addressed this with 04bcb8a8a6cfc02ce7c6eebd4088fac4e4ac902c and updated the PR description to handle it with this PR instead of a follow-up PR.

car-bridge/src/main.rs
```diff
@@ -29,58 +41,200 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+
+                        {
+                            let client = Arc::clone(&client);
+
+                            spawn(async move {
+                                if let Err(e) = client.publish(topic, message).await {
+                                    debug!("Error when publishing message: '{:?}'.", e);
+                                }
+                            });
+                        }
```
atifaziz, line 78: remove redundant scope:

```suggestion
                        let client = Arc::clone(&client);

                        spawn(async move {
                            if let Err(e) = client.publish(topic, message).await {
                                debug!("Error when publishing message: '{:?}'.", e);
                            }
                        });
```

car-bridge/src/main.rs
```diff
@@ -29,58 +41,198 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
 
     Ok(())
 }
+
+async fn handle_message(
+    chariott: &mut impl ChariottCommunication,
+    response_sender: Sender<(String, MessageBuilder)>,
+    message: MqttMessage,
+) {
+    async fn get_response(
```
atifaziz, line 118: Curious why you use a local function for something that's just used/called once?

bastbu: Yes, the idea is that for every fallible operation within `get_response` the action that we take is the same: we wrap the error message as an error response. For errors that happen outside of the `get_response` we either don't have the correlation data yet (and hence do not know where to return an error), or get an error that we can not recover from (which we log instead). The local function allows  us to handle all "error messages" once and in a unified way.

car-bridge/src/main.rs
```diff
@@ -29,58 +41,198 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+
+    let correlation_information = match message.get_correlation_information() {
+        Ok(cm) => cm,
+        Err(error) => {
+            debug!("Error when getting correlation information from message: '{error:?}'.");
+            return;
+        }
+    };
+
+    let response = match get_response(chariott, &message).await {
```
atifaziz, line 156: I think you can get rid of the local function `get_response` by just inlining it as a fallible async block:

```suggestion
    let correlation_information = match message.get_correlation_information() {
        Ok(cm) => cm,
        Err(error) => {
            debug!("Error when getting correlation information from message: '{error:?}'.");
            return;
        }
    };

    let response: Result<_, Box<dyn std::error::Error + Send + Sync>> = async {
        let fulfill_request: FulfillRequest = Message::decode(message.payload())?;

        let intent_enum = fulfill_request
            .intent
            .and_then(|intent| intent.intent)
            .ok_or_else(|| Error::new("Message does not contain an intent."))?;

        let response = match intent_enum {
            IntentEnum::Discover(_) => Err(Error::new("Discover is not supported.")),
            IntentEnum::Subscribe(_) => todo!(),
            _ => chariott
                .fulfill(fulfill_request.namespace, intent_enum)
                .await
                .map(|r| r.into_inner()),
        }?;

        let mut payload = vec![];
        response.encode(&mut payload)?;

        Ok(Response {
            payload,
            content_type: "application/x-proto+chariott.common.v1.FulfillResponse",
            is_error: false,
        })
    }
    .await;

    let response = match response {
```

It keeps the flow sequential: get correlation info or exit, get response or exit, etc.

car-bridge/src/main.rs
atifaziz: See [suggestion on inlining this](https://github.com/eclipse/chariott/pull/52/files#r1031475149).

car-bridge/src/main.rs
```diff
@@ -29,58 +41,196 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+    properties.push_string(PropertyCode::ContentType, response.content_type).unwrap();
+
+    properties
+        .push_string_pair(
+            PropertyCode::UserProperty,
+            "error",
+            match response.is_error {
+                true => "1",
+                false => "0",
+            },
```
atifaziz, line 185: Simpler as an `if`?

```suggestion
            if response.is_error { "1" } else { "0" },
```

car-bridge/src/main.rs
```diff
@@ -29,58 +41,193 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+                payload,
+                content_type: "application/x-proto+chariott.common.v1.Value",
+                is_error: true,
+            }
+        }
+    };
+
+    let mut properties = Properties::new();
+
+    properties.push_string(PropertyCode::ContentType, response.content_type).unwrap();
```
atifaziz, line 176: We're unwrapping/panicking for this and setting of the error property, but not for the correlation data property. To avoid this inconsistency, we should either panic on failure or trace and exit when setting any property.

car-bridge/src/main.rs
```diff
@@ -29,58 +41,193 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+            .ok_or_else(|| Error::new("No correlation data found on message."))?;
+
+        let response_topic = self
+            .properties()
+            .get_string(PropertyCode::ResponseTopic)
+            .ok_or_else(|| Error::new("No response topic found on message."))?;
+
+        Ok(CorrelationInformation { correlation_data, response_topic })
+    }
+}
```
atifaziz, line 233: nit: Since the extension only needs properties, it might be more general to just require that and default the rest of the implementation into the trait:

```suggestion
trait MqttExt {
    fn properties(&self) -> &Properties;

    fn get_correlation_information(&self) -> Result<CorrelationInformation, Error> {
        let correlation_data = self
            .properties()
            .get_binary(PropertyCode::CorrelationData)
            .ok_or_else(|| Error::new("No correlation data found on message."))?;

        let response_topic = self
            .properties()
            .get_string(PropertyCode::ResponseTopic)
            .ok_or_else(|| Error::new("No response topic found on message."))?;

        Ok(CorrelationInformation { correlation_data, response_topic })
    }
}

impl MqttExt for MqttMessage {
    fn properties(&self) -> &Properties {
        self.properties()
    }
}
```

car-bridge/src/messaging.rs
```diff
@@ -0,0 +1,168 @@
...
+#[async_trait]
+pub trait Subscriber {
+    type Message;
+    type Topic;
+
+    /// Subscribe to a topic.
+    async fn subscribe(
+        &mut self,
+        topic: String,
+    ) -> Result<BoxStream<'static, Self::Message>, Error>;
```
atifaziz, line 29: Why depend on Chariott's `Error`?

car-bridge/src/main.rs
bastbu: The difference that I wanted to express with the panic/error is that both the content type and the error user property are under the Car Bridge's control - if we suddenly fail to set those, it's a bug.

The correlation data is driven by the publisher, and hence it is not under the Car Bridge's control, which is why I was failing differently.

I used recoverable errors for consistency with 06418b39e423783d79a9e1eda3cfee15975a62e7, which also introduces some duplication in how we handle these errors. I would suggest to not attempt to fix this, at it will look differently in the follow-up PR #63.

car-bridge/src/messaging.rs
bastbu: We will need some error type. For now, I introduced an associated type to not require this as part of the trait definition with 2319d93fcb329ac9f71e8ae9c64ac74ba859d01e.

I kept reusing `chariott-common::error::Error` for the `MqttMessaging` implementation because I find it convenient. If I remove it, I would end up using `anyhow` (which is also just reusing another "common" error implementation) or creating an error type for the Car Bridge (which I don't think is worth it, as it would look identical to `chariott_common::error::Error`).

car-bridge/src/messaging.rs
```diff
@@ -0,0 +1,172 @@
...
+    type Topic = String;
+    type Error = Error;
+
+    /// Subscribes to a topic on the MQTT broker. Currently, this can only be
+    /// invoked once on `MqttMessaging`, due to the structure of the underlying
+    /// client.
+    async fn subscribe<'a>(
+        &'a mut self,
+        topic: String,
+    ) -> Result<BoxStream<'static, Self::Message>, Error> {
```
atifaziz, line 119: nit: Use `Self::` consistently:

```suggestion
    ) -> Result<BoxStream<'static, Self::Message>, Self::Error> {
```

car-bridge/src/messaging.rs
```diff
@@ -0,0 +1,172 @@
...
+}
+
+#[async_trait]
+impl Publisher for MqttMessaging {
+    type Message = MessageBuilder;
+    type Topic = String;
+    type Error = Error;
+
+    /// Publish a message to an MQTT broker.
+    async fn publish(&self, topic: Self::Topic, message: Self::Message) -> Result<(), Error> {
```
atifaziz, line 166: nit: Use `Self::` consistently:

```suggestion
    async fn publish(&self, topic: Self::Topic, message: Self::Message) -> Result<(), Self::Error> {
```

car-bridge/src/messaging.rs
atifaziz: It seems all your error are just string literals, so why not just use `&'static str` as your error type? Seems to be simple and works:

```diff
diff --git a/car-bridge/src/messaging.rs b/car-bridge/src/messaging.rs
index ecd3e513..5b54ebc6 100644
--- a/car-bridge/src/messaging.rs
+++ b/car-bridge/src/messaging.rs
@@ -6,10 +6,7 @@ use std::time::Duration;
 use async_channel::Receiver;
 use async_stream::stream;
 use async_trait::async_trait;
-use chariott_common::{
-    config::env,
-    error::{Error, ResultExt as _},
-};
+use chariott_common::config::env;
 use futures::{stream::BoxStream, StreamExt as _};
 use paho_mqtt::{
     AsyncClient, ConnectOptionsBuilder, CreateOptionsBuilder, Message, MessageBuilder,
@@ -58,7 +55,7 @@ impl MqttMessaging {
     /// If there was a persisted session before, messages may delivered before
     /// the `connect` returns. Refer to the `Subscriber` implementation for how
     /// to get access to the buffered messages.
-    pub async fn connect(client_id: String) -> Result<Self, Error> {
+    pub async fn connect(client_id: String) -> Result<Self, &'static str> {
         const BROKER_URL_ENV_NAME: &str = "BROKER_URL";
         const DEFAULT_BROKER_URL: &str = "tcp://localhost:1883";
         const MQTT_CLIENT_BUFFER_SIZE: usize = 200;
@@ -81,7 +78,7 @@ impl MqttMessaging {
                 .client_id(client_id)
                 .finalize(),
         )
-        .map_err_with("Failed to create MQTT client.")?;
+        .map_err(|_| "Failed to create MQTT client.")?;
 
         // Get the stream before connecting the client, as otherwise messages
         // may be lost. If the Car Bridge restarts while the number of
@@ -98,7 +95,7 @@ impl MqttMessaging {
                     .finalize(),
             )
             .await
-            .map_err_with("Could not connect to MQTT broker.")?;
+            .map_err(|_| "Could not connect to MQTT broker.")?;
 
         Ok(Self { client, receiver, is_subscribed: false })
     }
@@ -108,7 +105,7 @@ impl MqttMessaging {
 impl Subscriber for MqttMessaging {
     type Message = Message;
     type Topic = String;
-    type Error = Error;
+    type Error = &'static str;
 
     /// Subscribes to a topic on the MQTT broker. Currently, this can only be
     /// invoked once on `MqttMessaging`, due to the structure of the underlying
@@ -116,7 +113,7 @@ impl Subscriber for MqttMessaging {
     async fn subscribe<'a>(
         &'a mut self,
         topic: String,
-    ) -> Result<BoxStream<'static, Self::Message>, Error> {
+    ) -> Result<BoxStream<'static, Self::Message>, Self::Error> {
         // TODO: By broadcasting the events on the underlying `Receiver` and
         // filtering the broadcasted events by their topic name, we can support
         // multiple subscriptions. Since this is currently not needed, we do not
@@ -124,7 +121,7 @@ impl Subscriber for MqttMessaging {
         // instead.
 
         if self.is_subscribed {
-            return Err(Error::new("Already receiving messages. It is currently not possible to subscribe multiple times."));
+            return Err("Already receiving messages. It is currently not possible to subscribe multiple times.");
         }
 
         self.is_subscribed = true;
@@ -135,7 +132,7 @@ impl Subscriber for MqttMessaging {
         self.client
             .subscribe(topic.clone(), QOS_2)
             .await
-            .map_err_with("Could not subscribe to topic for receiving C2D messages.")?;
+            .map_err(|_| "Could not subscribe to topic for receiving C2D messages.")?;
 
         let mut receiver = self.receiver.clone();
 
@@ -160,13 +157,13 @@ impl Subscriber for MqttMessaging {
 impl Publisher for MqttMessaging {
     type Message = MessageBuilder;
     type Topic = String;
-    type Error = Error;
+    type Error = &'static str;
 
     /// Publish a message to an MQTT broker.
-    async fn publish(&self, topic: Self::Topic, message: Self::Message) -> Result<(), Error> {
+    async fn publish(&self, topic: Self::Topic, message: Self::Message) -> Result<(), Self::Error> {
         self.client
             .publish(message.topic(topic).finalize())
             .await
-            .map_err_with("Error when publishing a response.")
+            .map_err(|_| "Error when publishing a response.")
     }
 }
```

car-bridge/src/messaging.rs
```diff
@@ -0,0 +1,172 @@
...
+    /// If there was a persisted session before, messages may delivered before
+    /// the `connect` returns. Refer to the `Subscriber` implementation for how
+    /// to get access to the buffered messages.
+    pub async fn connect(client_id: String) -> Result<Self, Error> {
+        const BROKER_URL_ENV_NAME: &str = "BROKER_URL";
+        const DEFAULT_BROKER_URL: &str = "tcp://localhost:1883";
+        const MQTT_CLIENT_BUFFER_SIZE: usize = 200;
+
+        let host = env::<String>(BROKER_URL_ENV_NAME);
+        let host = host.as_deref().unwrap_or(DEFAULT_BROKER_URL);
```
atifaziz, line 67: These belong in `main.rs` and should really be passed to `connect` as arguments.

car-bridge/src/messaging.rs
bastbu: It does not have the same semantics. When we use `map_err_with` we keep the inner error, but provide additional context through the error message. When we use a `&'static str`, we discard all additional context that we may get from an inner error.

car-bridge/src/messaging.rs
bastbu: Done in 3db1b4234b2d95febfe8fc1f9648d139fce84059

car-bridge/src/main.rs
```diff
@@ -29,58 +44,203 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+        // backpressure and disconnect the client gracefully.
 
-    // Publish a phony message.
-    client.publish(Message::new(c2d_topic, "Example payload", QOS_2)).await?;
+        let cancellation_token = cancellation_token.child_token();
+        spawn(async move {
+            let client = Arc::new(client);
 
-    let cancellation_token = ctrl_c_cancellation();
+            loop {
```
atifaziz, line 67: Consider folding this into the loop and `select!` below, which should simplify things and get rid of one `spawn`.

car-bridge/src/main.rs
```diff
@@ -29,58 +44,203 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
-                    while let Err(err) = client.reconnect().await {
-                        info!("Trying to reconnect: {}.", err);
-                        sleep(Duration::from_secs(5)).await;
-                    }
-                }
+                let mut chariott = chariott.clone();
+                let response_sender = response_sender.clone();
+
+                spawn(async move {
+                    handle_message(&mut chariott, response_sender, message).await;
```
atifaziz, line 103: The setup here is hinting that `handle_message` ought to consume the cloned `choriott` rather than take a shared reference, since this is what's happening via the async block (or future) anyhow. It should then make the call here look as simple as:

```rust
spawn(handle_message(chariott, response_sender, message));
```

car-bridge/src/main.rs
bastbu: Done in c94b8b411264c9100b9b29ef6ef760cccef2688a

car-bridge/src/main.rs
bastbu: Done in 1ff909afe719ecc237a708aab7a20af7a2d9841d

atifaziz (approved):

car-bridge/src/messaging.rs
atifaziz: Ah, yes, I missed that in reading the code and had to refresh my memory of how `map_err_with` maintains the context.

PR #58, opened 2022-11-21
Cloud app for interacting via Car Bridge

### LINKED ISSUE #59: Enhance cloud app to interact via Car Bridge
## Description

In #50, a cloud app scaffolding was erected. The application simply and generally interacted with an MQTT broker, but should be changed to send proto messages to the Car Bridge to fulfill intents.

## Acceptance criteria

- [x] Application sends fulfillment requests to car topic.
- [x] Application parses corresponding fulfillment response and displays it.
- [x] Application gives up if a response is not received within a given period.
- [x] All intents are supported except those related to discovery.

atifaziz:
Closes #59.

## Motivation and Context

Add cloud-side application that exercises the MQTT and Car Bridge path to interact with the in-car applications.

## Description

This PR implements the cloud-side application as a .NET 6 console application in a similar spirit as [`charc`](https://github.com/eclipse/chariott/blob/6d2e9db2303b3d8b88ed62ec0aa8ce92026c160e/tools/charc.md) so that it can be used either interactive demos or for automation through a script.

The application can be launched with the following command-line options (none of which are required):

    Usage:
        CarBridgeCloudApp [--broker=<host>] [--vin=<vin>] [--timeout=<sec>] [--pretty-events]
        CarBridgeCloudApp -h | --help

    Options:
        -h --help        Show this screen.
        --broker=<host>  MQTT broker address [default: localhost].
        --vin=<vin>      VIN umber [default: 1].
        --timeout=<sec>  Timeout in seconds [default: 5].
        --pretty-events  Pretty print events JSON.

Once running, it takes commands from standard input that can be used to drive interactivity with the Car Bridge and in-car applications:

    ping
    set vin <vin>
    get vin
    get events.file
    inspect <namespace> <query>
    read <namespace> <key>
    write <namespace> <key> <value>
    invoke <namespace> <command> [<arg>...]
    subscribe <namespace> <source>...
    show topics
    show new events
    show value <value>
    (quit | exit)
    help

These are all described in the read-me document in the solution folder.


examples/applications/cloud-app/src/Program.cs
```diff
@@ -2,109 +2,482 @@
...
+                        }
+                        else
+                        {
+                            Console.Error.WriteLine($"Invalid value: {argValuePairs.First(v => v.Value is null).Arg}");
+                        }
+                        break;
+                    }
+                    case { CmdSubscribe: true, ArgNamespace: var ns, ArgSource: var sources }:
+                    {
+                        Debug.Assert(ns is not null);
```
bastbu, line 302: Should we consider creating the events file if it does not yet exist if we subscribe? When I tried tail the file when no event was published, I had to follow a specific order (publish inside the car, then tail) to get this done. It might easier for the developer if we `touch` the file, even if we received no events so far.

examples/applications/cloud-app/src/Program.cs
```diff
@@ -2,109 +2,482 @@
...
+        var options =
+            mqttFactory.CreateClientOptionsBuilder()
+                       .WithTcpServer(args.OptBroker)
+                       .WithProtocolVersion(MqttProtocolVersion.V500)
+                       .Build();
+        return mqttClient.ConnectAsync(options, cancellationToken);
+    });
 
-    await mqttClient.ConnectAsync(mqttClientOptions, CancellationToken.None);
     Console.Error.WriteLine("The MQTT client is connected.");
```
bastbu, line 54: [nit] should we write this to standard output instead?

examples/applications/cloud-app/src/Program.cs
```diff
@@ -2,109 +2,482 @@
...
+                          [NotNullWhen(true)] out Value.ValueOneofCase? kind,
+                          [NotNullWhen(true)] out Value? value)
 {
-    Console.Error.WriteLine(ex);
-    return 1;
+    input = input.Trim();
+    kind = default;
+    value = null;
+
+    if (input is "true")
```
bastbu, line 362: [nit] Should we be case-insensitive here?

examples/applications/cloud-app/src/Program.cs
```diff
@@ -2,109 +2,482 @@
...
+        from f in new DirectoryInfo(eventFilesDirPath).EnumerateFiles("*" + eventsFileExtension, new EnumerationOptions())
+        where f.CreationTime.Date < oldDate
+        select f;
+
+    foreach (var file in oldEventsFiles)
+        file.Delete();
+
+    var eventsFilePath = Path.Join(eventFilesDirPath, eventsTopic.Replace('/', '=') + eventsFileExtension);
+    var eventsFileReadPosition = 0L;
+    var eventsFileLock = new SemaphoreSlim(1);
```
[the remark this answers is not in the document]
atifaziz: Not strictly needed because the wait handle is never used, but let's err on the safe side in case the implementation changes in the future so committed with ae1dbf29e8a882e23f1058c6a694e82c88133c76.

examples/applications/cloud-app/src/Program.cs
atifaziz: This is for information purposes only so it's like logging, which generally goes to standard error.

examples/applications/cloud-app/src/Program.cs
atifaziz: Everything else is case-sensitive (like all commands and options) so it would be odd to make this an exception.

examples/applications/cloud-app/src/Program.cs
atifaziz: Good idea and done with cbe727f6f7142f715a88887a084a5757a4280e31.

bastbu (approved):

PR #60, opened 2022-11-23
Use `build_bindgen` feature to fix paho build problems

bastbu:
## Motivation and Context

Lately, the Rust CI for Paho MQTT has started failing.

## Description

As [recommended by Paho](https://github.com/eclipse/paho.mqtt.rust#using-the-crate), we use the `build_bindgen` feature to regenerate the bindings for our target. We also ensure that the required dependencies are installed when executing the Rust CI (which builds the Car Bridge).

### DIFF
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -34,8 +34,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Install protobuf-compiler
-        run: sudo apt-get install -y protobuf-compiler
+      - name: Install dependencies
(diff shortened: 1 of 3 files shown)


atifaziz (commented):
Wouldn't we need the same changes on the other workflows, like E2E, bench and LT?

bastbu:
> Wouldn't we need the same changes on the other workflows, like E2E, bench and LT?

When I checked LT, E2E and Benches they were not building the entire workspace, hence I did not add the installation of the dependencies there. In case we end up building the entire workspace as part of these CIs, it could make sense to install the dependencies already, though.

atifaziz (approved):
> When I checked LT, E2E and Benches they were not building the entire workspace, hence I did not add the installation of the dependencies there.

Makes sense. Thanks!

PR #61, opened 2022-11-23
Install dependencies required for Paho MQTT in Devcontainer

bastbu:
## Description

With this PR we install the dependencies required for building the Car Bridge inside the Devcontainer. This is an omission from #60.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -7,6 +7,7 @@
 ARG VARIANT="bullseye"
 FROM mcr.microsoft.com/vscode/devcontainers/rust:0-${VARIANT}
 
+# `libssl-dev`, `build-essential` and `clang` are required for Paho MQTT
 RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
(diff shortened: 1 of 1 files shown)


.devcontainer/Dockerfile
```diff
@@ -8,7 +8,7 @@ ARG VARIANT="bullseye"
 FROM mcr.microsoft.com/vscode/devcontainers/rust:0-${VARIANT}
 
 RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
-    && apt-get -y install --no-install-recommends git-lfs cmake protobuf-compiler libprotobuf-dev
+    && apt-get -y install --no-install-recommends git-lfs cmake protobuf-compiler libprotobuf-dev libssl-dev build-essential cmake clang
```
atifaziz, line 11: `cmake` is mentioned twice:

```suggestion
    && apt-get -y install --no-install-recommends git-lfs cmake protobuf-compiler libprotobuf-dev libssl-dev build-essential clang
```

nit: Someone will forget, so should we comment here why these new dependencies are needed? This way, they can remove them if ever Maho MQTT is no longer being used.

atifaziz (approved):

PR #63, opened 2022-11-24
Streaming events from the car to the cloud

### LINKED ISSUE #62: Stream data via the Car Bridge
## Description

The Car Bridge should be able to fulfill streaming-related intents.

## Acceptance criteria

- [x] Car Bridge fulfills subscription-related requests inside the car on-behalf of cloud applications
- [x] Listening management: If multiple sources are subscribed to within the car that originate from the same service, we only want to listen once on the streaming endpoint.
- [x] Subscription management: If multiple cloud applications subscribe to the same sources in Chariott, we only want to subscribe once to the underlying source

bastbu:
Closes #62 

## Motivation and Context

With this PR the Car Bridge can fulfill streaming-related intents. Depends on #52.

## Description

With this PR we introduce support for the `Subscribe` intent inside the Car Bridge. The Car Bridge will manage subscriptions to the target providers: for example, it will only open at maximum one listening channel for every provider. Also, it will never publish the same event twice to the same topic, nor subscribe to the same event twice inside the car.

The cloud application must use the `chariott.common.v1.SubscribeIntent.channel_id` to drive the topic on which the Car Bridge will publish events. If multiple applications use the same `channel_id`, events will only be published once.

Unregistering subscriptions is not done as part of this PR, but will be part of a future PR. Hence, the need for cancelling links, routes, subscriptions or listening actions is not considered.


examples/applications/lt-provider/Cargo.toml
```diff
@@ -7,7 +7,7 @@ license = "MIT"
 [dependencies]
 chariott-common = { workspace = true }
 chariott-proto = { workspace = true }
-examples-common = { path = "../../common/" }
+examples-common = { workspace = true }
```
atifaziz, line 10: Is this an unrelated change? If so, why not roll this together with replacing all other paths with the workspace flag?

bastbu: Not strictly unrelated. Since with this PR I needed to take a dependency on `examples-common`, I wanted to do this without causing more duplication. Hence I used workspace inheritance for the dependency that I was going to depend on. To stay consistent, this had to be changed for the projects that already had a dependency on said dependency.

car-bridge/src/streaming.rs
```diff
@@ -0,0 +1,222 @@
...
+use ess::{EventSubSystem, NotReadingEvents};
+use examples_common::chariott::api::{Chariott, ChariottCommunicationExt as _};
+use prost::Message as _;
+use tokio::spawn;
+use tokio_stream::StreamExt as _;
+use tracing::{debug, warn};
+
+use crate::{Message, ResponseSender};
+
+/// Identifies a namespace
```
atifaziz, line 20: nit: Self-explanatory, no?

```suggestion
```

car-bridge/src/streaming.rs
```diff
@@ -0,0 +1,222 @@
...
+type Namespace = String;
+/// Identifies a topic
```
atifaziz, line 22: nit: Self-explanatory, no?

```suggestion
```

car-bridge/src/streaming.rs
```diff
@@ -0,0 +1,222 @@
...
+type Topic = String;
+/// Identifies a subscription source (a.k.a. _key_).
+type Source = String;
```
atifaziz, line 25: nit: Want to call it `SubscriptionSource` so the comment isn't needed?

car-bridge/src/streaming.rs
```diff
@@ -0,0 +1,222 @@
...
+
+        spawn(async move {
+            while let Some(e) = topic_stream.next().await {
+                let mut buffer = vec![];
+
+                if let Err(err) = e.encode(&mut buffer) {
+                    debug!("Failed to encode event: '{err:?}'.");
+                }
+
+                response_sender.send(Message::Event(buffer, topic.clone())).await;
```
atifaziz, line 197: I don't think you want to send a potentially corrupted/partial message if encoding fails:

```suggestion
                let mut buffer = Vec::with_capacity(e.encoded_len());
                match e.encode(&mut buffer) {
                    Ok(_) => response_sender.send(Message::Event(buffer, topic.clone())).await,
                    Err(err) => debug!("Failed to encode event: '{err:?}'."),
                };
```

car-bridge/src/streaming.rs
bastbu: done in 7a0eb903e02d9c3457de4649e8f5cbdbe3b087e7

atifaziz (approved):

PR #65, opened 2022-11-28
Attempt to complete all in-flight tasks when shutting down Car Bridge

bastbu:
## Motivation and Context

When shutting down the Car Bridge we want to finish publishing as many messages as possible. Since we detach the message handling and publishing of tasks, it is possible that if the Car Bridge shuts down, there are in-flight messages that were not yet processed.

## Description

With this PR we attempt to finish processing these in-flight messages. The approach is based on an `mpsc` channel as described in the [Tokio docs](https://tokio.rs/tokio/topics/shutdown).

### DIFF
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -9,6 +9,7 @@ on:
       - main
       - cloud-bridge
     paths:
+      - 'car-bridge/**'
       - 'common/**'
       - 'src/**'
       - 'tests/**'
--- a/car-bridge/src/drainage.rs
+++ b/car-bridge/src/drainage.rs
@@ -1,3 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
 use futures::Future;
 use tokio::{
     spawn,
--- a/car-bridge/src/drainage.rs
+++ b/car-bridge/src/drainage.rs
@@ -2,11 +2,7 @@
 // Licensed under the MIT license.
 
 use futures::Future;
-use tokio::{
-    spawn,
-    sync::mpsc::{self, Receiver, Sender},
-    task::JoinHandle,
-};
+use tokio::sync::mpsc::{self, Receiver, Sender};
 
 /// Spawn tasks on the `Drainage` to ensure that the task completed when
 /// shutting down.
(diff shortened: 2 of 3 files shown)


car-bridge/src/drainage.rs
```diff
@@ -0,0 +1,72 @@
...
+            drop(sender);
+            result
+        })
+    }
+
+    /// Waits for all tasks that were spawned via this drainage to finish.
+    pub async fn drain(mut self) {
+        drop(self.drained_sender);
+        // As soon as every sender is dropped, the channel is closed and the
+        // receiver receives a message. Based on:
```
atifaziz, line 44: I think you mean to say that the reception fails rather than getting a message?

```suggestion
        // reception fails. Based on:
```

car-bridge/src/main.rs
```diff
@@ -66,7 +69,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
                     break;
                 };
 
-                spawn(handle_message(chariott.clone(), response_sender.clone(), message));
+                drainage.spawn(handle_message(chariott.clone(), response_sender.clone(), message));
```
atifaziz: There's not much point in making this part of the drainage since any responses from the `handle_message` will never get published since this loop is broken by the cancellation token. Ideally, these are aborted too by the cancellation token.

.github/workflows/rust-ci.yml
```diff
@@ -9,6 +9,7 @@ on:
       - main
       - cloud-bridge
     paths:
+      - 'car-bridge/**'
```
atifaziz, line 12: Isn't this unrelated to this PR?

car-bridge/src/main.rs
bastbu: Agreed, removed with f586d8da73a7b381b5e33b64dafe85cd6d19ebea.

.github/workflows/rust-ci.yml
bastbu: Unrelated, yes, but it became apparent with this PR as now we have two unit tests as part of the Car Bridge. I can PR it separately, or we can keep it as it is. No strong preference.

atifaziz (approved):
Apart from a couple of comment, LGTM!

.github/workflows/rust-ci.yml
atifaziz: Let's stick with consistently submitting unrelated changes as separate PRs that are easier to track.

bastbu: #66

car-bridge/src/drainage.rs
```diff
@@ -0,0 +1,72 @@
...
+
+impl Drainage {
+    pub fn new() -> Self {
+        let (drained_sender, drained_receiver) = mpsc::channel(1);
+        Self { drained_receiver, drained_sender }
+    }
+
+    /// Spawns a task while allowing to wait for it to complete when calling
+    /// `drain`.
+    pub fn spawn<F>(&self, f: F) -> JoinHandle<F::Output>
```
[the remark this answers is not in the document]
bastbu: Valid point, addressed in 644ecd3ea257378eda48702cbfcf8f72904d1b22.

PR #66, opened 2022-11-28
Trigger Rust CI when `car-bridge` changed

bastbu:
## Description

We want to run the Rust CI when the `car-bridge` directory contains changes.


atifaziz (approved):
Thanks!

PR #67, opened 2022-11-28
Support cancellation of `handle_message`

bastbu:
## Description

With this PR we support cancelling in-flight `handle_message` tasks. When cancellation is requested, all in-flight fulfillment `Future`s against Chariott are dropped and instead an error is returned. When shutting down, all errors are (attempted to be) delivered to the MQTT broker to notify applications that the operation was cancelled.

### DIFF
--- a/car-bridge/src/main.rs
+++ b/car-bridge/src/main.rs
@@ -60,13 +60,14 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     let cancellation_token = ctrl_c_cancellation();
     let drainage = Drainage::new();
-    let mut shutdown_handled = false;
+    // Handle cancellation exactly once.
+    let mut cancellation_handled = false;
 
     let (mut response_sender, mut response_receiver) = mpsc::channel(PUBLISH_BUFFER);
 
     loop {
         select! {
-            message = messages.next(), if !shutdown_handled => {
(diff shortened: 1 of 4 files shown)


car-bridge/src/main.rs
```diff
@@ -61,19 +62,24 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let drainage = Drainage::new();
 
     let (response_sender, mut response_receiver) = mpsc::channel(PUBLISH_BUFFER);
+    let mut response_sender = Some(response_sender);
 
     loop {
         select! {
-            message = messages.next() => {
+            message = messages.next(), if response_sender.is_some() => {
```
atifaziz, line 69: nit: avoid `unwrap` by moving the guard inside:

```rust
message = messages.next() => {
    if let Some(ref response_sender) = response_sender {
        let Some(message) = message else {
            break;
        };
        spawn(handle_message(chariott.clone(), response_sender.clone(), message, cancellation_token.child_token()));
    }
}
```

I don't think there's anything to be gained by the guard on the `select!` arm since there isn't another `message = messages.next()` to disambiguate with.

bastbu: Addressed in f5629e7061003a686990331b34cbde8ca20c7b13

PR #79, opened 2023-01-13
fix: RUSTSEC-2023-0001

dariuszparys:
This will bump tokio to 1.23.1 to fix the current security audit warning RUSTSEC-2023-0001

See https://rustsec.org/advisories/RUSTSEC-2023-0001


ladatz:
this was addressed in this PR: https://github.com/eclipse-chariott/chariott/pull/155#pullrequestreview-1527206676

PR #93, opened 2023-03-07
Added an invoke command example provider

devkelley:
## Motivation and Context
This PR is needed to provide a simple example of an invoke intent implementation on the provider side.

## Description
This code change adds a simple example of how a provider would be set up to handle an invoke call through chariott. It also provides a readme file that gives some sample grpcurl commands to test the invoke call.

**PR COMMIT MESSAGE**

<type>docs: example code showing simple invoke intent implementation</description>


examples/applications/invoke-command/src/main.rs
```diff
@@ -0,0 +1,43 @@
...
+use crate::chariott_provider::ChariottProvider;
+
+chariott::provider::main!(wain);
+
+async fn wain() -> Result<(), Error> {
+    let (url, socket_address) = chariott::provider::register(
+        "sdv.invoke.controller",
+        "0.0.1",
+        "sdv.invoke.controller",
+        [Intent::Invoke, Intent::Discover],
```
ladatz, line 28: Consider putting these in the order of the enum: https://github.com/eclipse/chariott/blob/b0d2add377f1740f26c52bfb4e234ce415b91b81/src/chariott_grpc.rs#:~:text=const%20INTENT_MAPPING_DISCOVER%3A,%3D%205%3B

devkelley: updated

Issue #104, opened 2023-03-22
[BUG] Building Arm64 dev container fails


benmordaunt:
While building the dev container using `Dockerfile.arm64`, the following error is encountered:

```
   Compiling tracing-log v0.1.3
   Compiling sharded-slab v0.1.4
   Compiling ess v0.1.0 (/sdv/ess)
   Compiling form_urlencoded v1.1.0
   Compiling chariott v0.1.0 (/sdv)
   Compiling uuid v1.3.0
   Compiling thread_local v1.1.7
   Compiling smallvec v1.10.0
   Compiling tracing-subscriber v0.3.16
   Compiling chariott-common v0.1.0 (/sdv/common)
   Compiling url v2.3.1
   Compiling async-recursion v1.0.4
error[E0658]: attributes on expressions are experimental
   --> src/chariott_grpc.rs:144:9
    |
144 |         #[cfg(test)]
    |         ^^^^^^^^^^^^
    |
    = note: see issue #15701 <https://github.com/rust-lang/rust/issues/15701> for more information

error: removing an expression is not supported in this position
   --> src/chariott_grpc.rs:144:9
    |
144 |         #[cfg(test)]
    |         ^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0658`.
error: could not compile `chariott` due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
The command '/bin/sh -c cargo build --release --target=aarch64-unknown-linux-musl' returned a non-zero code: 101
```

benmordaunt:
Also fixed by #105

PR #105, opened 2023-03-23
Fix for docker build failure

### LINKED ISSUE #103: [BUG] Issue building Dockerfile in Chariott
There is currently an issue building Chariott when building a Dockerfile. This issue is not reproducible when directly running the cargo build command for Chariott.

## Expected Behavior

First, Dockerfiles in the repo should be able to be built without errors. Secondly, if there is an issue building the code, it should be reproducible both in the Docker build and when directly building the code with cargo.

## Current Behavior

Chariott can be built directly with cargo without any issues. However, when a Dockerfile is attempted to be built, using the same cargo command, the build fails with an error.

## Steps to Reproduce

1. run `rustup target add x86_64-unknown-linux-musl` to add target.
2. run `cargo build --release --target=x86_64-unknown-linux-musl` in the root of the repo, and see that it completes without error.
3. then, attempt to build a Dockerfile with `docker build -t chariott -f ./Dockerfile . ` in the root of the repo. This should result in an error, of type `error[E0658]: attributes on expressions are experimental`.

## Context (Environment)

Environment: WSL 2
cargo version: 1.65.0 (4bc8f24d3 2022-10-20)
Docker version: 23.0.1, build a5ee5b1
rust compiler (rustc) version: 1.65.0 (897e37553 2022-11-02) 

## Logs

> 154.0    Compiling chariott-common v0.1.0 (/sdv/common)
> 154.7 error[E0658]: attributes on expressions are experimental
> 154.7       --> src/chariott_grpc.rs:144:9
> 154.7        |
> 154.7 144 |         #[cfg(test)]
> 154.7        |         ^^^^^^^^^^^^
> 154.7        |
> 154.7        = note: see issue #15701 <https://github.com/rust-lang/rust/issues/15701> for more information
> 154.7 
> 154.7 error: removing an expression is not supported in this position
> 154.7       --> src/chariott_grpc.rs:144:9
> 154.7        |
> 154.7 144 |         #[cfg(test)]
> 154.7        |         ^^^^^^^^^^^^
> 154.7 
> 155.0 For more information about this error, try `rustc --explain E0658`.
> 155.0 error: could not compile `chariott` due to 2 previous errors

devkelley:
This provides a quick fix for an issue that causes a dockerfile build to fail when building Chariott with cargo. This issue is not reproducible when directly calling the cargo file.

Fixes #103 and #104

## Motivation and Context

The change is required because dockerfiles cannot currently be built in Chariott.

## Description

1. Added `let` in front of a line that is used when testing to suppress a dead code warning.
2. Added to a command that gets a rust version from a file when building a container for a git action.

### DIFF
--- a/src/chariott_grpc.rs
+++ b/src/chariott_grpc.rs
@@ -142,7 +142,9 @@ impl<T: Observer + Send + Sync + 'static> ChariottService for ChariottServer<T> 
         #[cfg(not(test))]
         let broker = &self.broker;
         #[cfg(test)]
-        let broker = tests::MockBroker::new(self.broker.clone());
+        let _ = self.broker; // Suppress dead code warning when test feature is active.
+        #[cfg(test)]
+        let broker = tests::MockBroker;
 
         let binding =
             broker.resolve(&config).ok_or_else(|| Status::not_found("No provider found."))?;
@@ -404,11 +406,6 @@ mod tests {
 
     impl MockBroker {
         const RETURN_VALUE: i32 = 10;
-
-        pub fn new(broker: IntentBroker) -> Self {
-            _ = broker; // Suppress dead code warning when test feature is active.
-            Self
-        }
 
         pub fn resolve(&self, _: &IntentConfiguration) -> Option<RuntimeBinding<GrpcProvider>> {
             Some(RuntimeBinding::Test(TestBinding::new(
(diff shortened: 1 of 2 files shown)


src/chariott_grpc.rs
```diff
@@ -142,7 +142,7 @@ impl<T: Observer + Send + Sync + 'static> ChariottService for ChariottServer<T>
         #[cfg(not(test))]
         let broker = &self.broker;
         #[cfg(test)]
-        _ = self.broker; // Suppress dead code warning when test feature is active.
+        let _ = self.broker; // Suppress dead code warning when test feature is active.
```
devkelley, line 145: I based this change off a conversation found here: https://github.com/rust-lang/rust/issues/59144

Issue #98, opened 2023-03-11
[BUG] CI checks are failing


ladatz:
## Expected Behavior
When a PR is opened with no code changes, the CI checks should pass.

## Current Behavior
CI checks are currently failing due to some deprecated actions, some broken links in documentation, and some 

## Steps to Reproduce

1. Create a new PR
2. Run the workflow ci checks

## Logs

Some examples: 
```
FILE: ./docs/adr/0001-service-discovery.md
https://cdn.vector.com/cms/content/events/2019/VH/VIC2019/Track_1_5_Glancing_Ethernet_communication_for_Automotive_4.pdf

  ERROR: 1 dead links found!

  19 links checked.
```

`Diff in /home/runner/work/chariott/chariott/examples/applications/cloud-object-detection/src/chariott_provider.rs at line 1:
 /// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 // SPDX-License-Identifier: MIT
-`

PR #112, opened 2023-04-04
Fix static code analysis and doc link errors

devkelley:
Partial fix for #98 

## Motivation and Context

This resolves current code static analysis errors and removes dead/inaccessible links in old documentation.

## Description

Fixed static analysis code errors throughout the solution. This included recent and older code. Also removed some links that either no longer exist or are inaccessible. These links are in older documentation that has been superseded and thus are not critical to maintain.

### DIFF
--- a/.markdownlinkcheck.jsonc
+++ b/.markdownlinkcheck.jsonc
@@ -1,4 +1,12 @@
 {
+  "httpHeaders": [
+    {
+      "urls": ["https://docs.github.com/"],
+      "headers": {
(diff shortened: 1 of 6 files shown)


devkelley:
Note the Lint and Check Markdown workflow fails because of a recent issue introduced in the new version related to the service not taking in the config file. The current link failures are only to localhost, which is ignored in the config. The workflow will pass once the issue is fixed. See issue: https://github.com/tcort/markdown-link-check/issues/246

PR #141, opened 2023-06-21
Simplify Service Discovery by decoupling the service registry from intent brokering

### LINKED ISSUE #134: [PBI] Simplify Service Discovery Process
## Description
Today, for an application to be a provider, it must implement the Provider interface and have a grpc endpoint to communicate on. This can be a lot of code required for an application that only wishes to expose an endpoint. Issue #117 is a great example of this. It would be better if Service Discovery had less requirements on a provider. Providers and Consuming applications should only need to call Chariott APIs, not implement any logic to be used.

## Acceptance criteria
- [ ] "Discover" is a simple API call
- [ ] "Providers" are not required to implement an interface

ladatz:
Closes #134 

## Motivation and Context

Today, Service Discovery and intent brokering are tightly coupled. Separating them allows simplification of providers. With this change, providers will only need to communicate with Chariott through clients, rather than needing to implement a specific gRPC interface. This enables simpler execution of the direct application-application communication model. We've received feedback from the community that people using Chariott for Service Discovery would prefer this simpler approach. 

## Description

Created a new service registry implementation which is completely separate from the intent brokering. This is the first iteration of the new registry, and features will be added over time, according to priority and community feedback.
<!--


service_discovery/samples/simple-discovery/provider/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "provider"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
jorchiu, line 6: Please add the license header, and add `license = "MIT"` under `edition = "2021"`

You can remove this comment.

service_discovery/core/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "service_discovery"
```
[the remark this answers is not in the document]
jorchiu: Please add the license header, and add `license = "MIT"` under `edition = "2021"`

service_discovery/samples/simple-discovery/provider/src/main.rs
```diff
@@ -0,0 +1,78 @@
...
+            EnvFilter::builder()
+                .with_default_directive(tracing::Level::INFO.into())
+                .from_env_lossy(),
+        )
+        .finish();
+
+    collector.init();
+
+    // Intitialize addresses for provider and chariott communication.
+    let provider_url_str = format!("http://{}", HELLO_WORLD_ENDPOINT);
```
jorchiu, line 49: `let provider_url_str = format!("http://{HELLO_WORLD_ENDPOINT}");`

service_discovery/proto/chariott/v1/chariott_registry.proto
```diff
@@ -0,0 +1,98 @@
...
+  rpc Inspect(InspectRequest) returns (InspectResponse) {}
+}
+
+// Represenation of a service including all of its metadata that the registry stores
+message ServiceMetadata {
+  string namespace = 1;
+  string name = 2;
+  string version = 3;
+  string uri = 4;
+  string communication_kind = 5;
```
jorchiu, line 36: Please consider making this an enum.

service_discovery/core/src/main.rs
```diff
@@ -0,0 +1,52 @@
...
+use std::sync::Arc;
+use tonic::transport::Server;
+use tracing::{debug, info};
+use tracing_subscriber::util::SubscriberInitExt;
+use tracing_subscriber::EnvFilter;
+
+mod registry_impl;
+
+/// Endpoint for the chariott service registry
+const CHARIOTT_SERVICE_REGISTRY_ADDR: &str = "0.0.0.0:50000";
```
liviuft, line 27: chariott is the code name; what about leaving it out of the code itself and just name each class based on its functionality

service_discovery/proto/build.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+// SPDX-License-Identifier: MIT
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    tonic_build::compile_protos("../proto/chariott/v1/chariott_registry.proto")?;
```
liviuft, line 6: same here in terms of using the code name of the project

service_discovery/samples/simple-discovery/consumer/src/main.rs
```diff
@@ -0,0 +1,68 @@
...
+        version: String::from("1.0.0.0"),
+    });
+
+    // Discover the simple provider service
+    let service_option =
+        registry_client.discover_service(discover_request).await?.into_inner().service;
+    match service_option {
+        Some(service) => {
+            info!("Discovered service {:?}", service);
+            if service.communication_kind != *"grpc+proto"
```
jorchiu, line 53: Please consider making communication_kind an enum, and then you can use these 2 crates: `strum` (https://docs.rs/strum/latest/strum/)  and `strum_macros`.

Specifically, you may want to use the EnumString attribute from the strum_macros crate (https://docs.rs/strum/latest/strum/derive.EnumString.html), which allows you to match a String to an enum variant

service_discovery/proto/build.rs
```diff
@@ -0,0 +1,9 @@
...
+    tonic_build::compile_protos("../proto/samples/v1/hello_world_service.proto")?;
```
liviuft, line 7: what is the link between this samples folder and the one under the service_discovery folder? can we have all sample under that folder?

service_discovery/proto/chariott/v1/chariott_registry.proto
```diff
@@ -0,0 +1,98 @@
...
+// The entry point for the Registry gRPC Service.
+service Registry {
+  // Register, or add a service to the registry
+  rpc Register(RegisterRequest) returns (RegisterResponse) {}
+
+  // Unregister, or remove a service from the registry
+  rpc Unregister(UnregisterRequest) returns (UnregisterResponse) {}
+
+  // Discover, or retrieve the metadata for a single service given its fully qualified name
+  rpc DiscoverService(DiscoverServiceRequest) returns (DiscoverServiceResponse) {}
```
liviuft, line 21: what about naming it just Discover; as this is a registry of services it is implied that it is about discovering services; for example we do not have RegisterService

service_discovery/proto/chariott/v1/chariott_registry.proto
```diff
@@ -0,0 +1,98 @@
...
+
+  // Discover a list of services given their namespace
+  rpc DiscoverByNamespace(DiscoverByNamespaceRequest) returns (DiscoverByNamespaceResponse) {}
+
+  // Inspect, or retrieve all contents of the service registry
+  rpc Inspect(InspectRequest) returns (InspectResponse) {}
```
liviuft, line 27: what about calling this just List?

jorchiu (commented):
Please take a look at my comments.

service_discovery/proto/chariott/v1/chariott_registry.proto
```diff
@@ -0,0 +1,98 @@
...
+  string namespace = 1;
+  string name = 2;
+  string version = 3;
+  string uri = 4;
+  string communication_kind = 5;
+  string communication_reference = 6;
+}
+
+// Status of a registration operation
+enum RegistrationStatus {
```
liviuft, line 41: what about just failing if trying to register something already registered? if we have requirements about doing update we can add that option in the future, for example through ways of specifying if one wants to overwrite and so on during the Register operation

service_discovery/core/src/registry_impl.rs
```diff
@@ -0,0 +1,457 @@
...
+use std::collections::HashMap;
+use std::sync::Arc;
+use std::vec::Vec;
+use tonic::{Request, Response, Status};
+use tracing::{info, warn};
+
+/// Base structure for the registry gRPC service
+#[derive(Clone, Debug)]
+pub struct RegistryImpl {
+    pub registry_map: Arc<RwLock<HashMap<ServiceIdentifiers, ServiceMetadata>>>,
```
liviuft, line 26: do we need to expose this externally? we can have specific methods on the registry for add/remove/get that, internally, access this specific structure

service_discovery/core/src/registry_impl.rs
```diff
@@ -0,0 +1,457 @@
...
+};
+use std::collections::HashMap;
+use std::sync::Arc;
+use std::vec::Vec;
+use tonic::{Request, Response, Status};
+use tracing::{info, warn};
+
+/// Base structure for the registry gRPC service
+#[derive(Clone, Debug)]
+pub struct RegistryImpl {
```
liviuft, line 25: what about calling this ServiceRegistry instead of just Registry?

service_discovery/core/src/registry_impl.rs
```diff
@@ -0,0 +1,457 @@
...
+            namespace: service_to_register.namespace.clone(),
+            name: service_to_register.name.clone(),
+            version: service_to_register.version.clone(),
+        };
+        info!("Received a register request for: {:?}", service_identifiers);
+
+        let registration_status;
+        // This block controls the lifetime of the lock.
+        {
+            let mut lock: RwLockWriteGuard<HashMap<ServiceIdentifiers, ServiceMetadata>> =
```
wilyle, line 56: consider omitting this rather verbose type annotation unless it is required. I find that these mutex types make the code harder to read because they're not easy to parse.

(this appears in a bunch of places in this PR)

service_discovery/proto/chariott/v1/chariott_registry.proto
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// Represenation of a service including all of its metadata that the registry stores
+message ServiceMetadata {
+  string namespace = 1;
+  string name = 2;
+  string version = 3;
+  string uri = 4;
+  string communication_kind = 5;
+  string communication_reference = 6;
```
liviuft, line 37: can we add some samples for what the two communication_* fields are?

service_discovery/proto/chariott/v1/chariott_registry.proto
```diff
@@ -0,0 +1,98 @@
...
+  rpc DiscoverService(DiscoverServiceRequest) returns (DiscoverServiceResponse) {}
+
+  // Discover a list of services given their namespace
+  rpc DiscoverByNamespace(DiscoverByNamespaceRequest) returns (DiscoverByNamespaceResponse) {}
+
+  // Inspect, or retrieve all contents of the service registry
+  rpc Inspect(InspectRequest) returns (InspectResponse) {}
+}
+
+// Represenation of a service including all of its metadata that the registry stores
```
liviuft, line 30: typo

service_discovery/core/src/registry_impl.rs
```diff
@@ -0,0 +1,457 @@
...
+                self.registry_map.write();
+            match lock.get(&service_identifiers) {
+                Some(_) => {
+                    lock.insert(service_identifiers.clone(), service_to_register.clone());
```
wilyle, line 60: because you make the same call to `lock.insert` twice, consider making this call after the match block instead

also, is the intent here to just completely discard the old entry?

service_discovery/core/src/registry_impl.rs
```diff
@@ -0,0 +1,457 @@
...
+                    registration_status = RegistrationStatus::Updated;
+                    info!("Updated the service entry in Chariott service registry; overwrote previous entry: {:?}", service_to_register);
```
wilyle, line 62: is this supposed to log the old entry or the new entry? the wording leads me to think it's the old one, but the value being logged is the new one

service_discovery/core/src/registry_impl.rs
```diff
@@ -0,0 +1,457 @@
...
+                    let not_found_message = format!(
+                        "Unable to remove service from registry: {:?}",
+                        service_identifiers
+                    );
+                    warn!(not_found_message); //namespace: {0}, name: {1}, version: {2}")
+                    return Err(Status::not_found(not_found_message));
+                }
+            };
+        }
+        Ok(Response::new(UnregisterResponse {}))
```
wilyle, line 120: I would move this to the end of the `Some` arm of the match and omit the `return` in front of `Err`. So it would have this structure:

```
match lock.remove(...) {
    Some(...) => {
        info!(...);
        Ok(Response::new(UnregisterResponse {}))
    }
    None => {
        warn!(...);
        Err(Status::not_found(not_found_message))
    }
}
```

service_discovery/core/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
```
jorchiu, line 1: Please consider adding documentation to this folder.

service_discovery/core/src/registry_impl.rs
```diff
@@ -0,0 +1,457 @@
...
+        request: Request<DiscoverByNamespaceRequest>,
+    ) -> Result<Response<DiscoverByNamespaceResponse>, Status> {
+        let namespace = request.into_inner().namespace;
+        let mut service_list: Vec<ServiceMetadata> = Vec::new();
+
+        // This block controls the lifetime of the lock.
+        {
+            let lock: RwLockReadGuard<HashMap<ServiceIdentifiers, ServiceMetadata>> =
+                self.registry_map.read();
+            for (service_identifier, service_metadata) in lock.iter() {
```
wilyle, line 142: consider using an iterator here:

```
let service_list = {
    let lock = self.registry_map.read();
    lock
        .iter()
        .filter_map(|(service_identifier, service_metadata)| {
            if service_identifier.namespace == namespace {
                Some(service_metadata.clone())
            } else {
                None
            }
        })
        .collect()
}
```

Cargo.toml
```diff
@@ -15,6 +15,9 @@ members = [
     "examples/common",
     "keyvalue",
     "proto.rs",
+    "service_discovery/core",
+    "service_discovery/samples/simple-discovery/consumer",
+    "service_discovery/samples/simple-discovery/provider"
 ]
 
 exclude = []
```
jorchiu, line 23: Is this exclude field needed?

service_discovery/core/Cargo.toml
ladatz: Agreed, I definitely will. I am weighing the option of including it in this PR or a follow-up. If I decide to do it in a follow-up, I will link a GitHub Issue for it to this comment

service_discovery/core/src/registry_impl.rs
```diff
@@ -0,0 +1,457 @@
...
+            self.registry_map.read();
+        let services_list = lock.values().cloned().collect();
+        let inspect_response = InspectResponse { services: services_list };
+        Ok(Response::new(inspect_response))
+    }
+}
+
+/// Identifiers for a given service.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct ServiceIdentifiers {
```
liviuft, line 225: we calls this the plural, but it represents a single service; what about calling it the singular?

service_discovery/proto/samples/v1/hello_world_service.proto
```diff
@@ -0,0 +1,29 @@
...
+//
+// The protobuf definitions for the Hello World service sample
+
+syntax = "proto3";
+package hello_world;
+
+// The service entry point to the Hello World service. This simple service only has one method to
+// show the basic flow of applications calling one another
+service HelloWorld {
+    // Method which just logs and returns a message "Hello, {input string}" when it is called
```
ashbeitz, line 15: Please keep your indentation consistent.  You use 4 spaces here and 2 spaces later in the file.

service_discovery/samples/simple-discovery/consumer/src/main.rs
```diff
@@ -0,0 +1,68 @@
...
+use proto_servicediscovery::chariott_registry::v1::DiscoverServiceRequest;
+use proto_servicediscovery::hello_world::v1::hello_world_client::HelloWorldClient;
+use proto_servicediscovery::hello_world::v1::HelloRequest;
+use tonic::Request;
+use tracing::info;
+use tracing_subscriber::util::SubscriberInitExt;
+use tracing_subscriber::EnvFilter;
+
+/// URL for the chariott service registry
+const CHARIOTT_SERVICE_REGISTRY_URL: &str = "http://0.0.0.0:50000";
```
ashbeitz, line 24: Please move this to a config file.

service_discovery/samples/simple-discovery/consumer/src/main.rs
ashbeitz, line 53: Please make "grpc+proto" a constant.

service_discovery/samples/simple-discovery/provider/src/main.rs
```diff
@@ -0,0 +1,78 @@
...
+use tonic::transport::Server;
+use tonic::Request;
+use tracing::info;
+use tracing_subscriber::util::SubscriberInitExt;
+use tracing_subscriber::EnvFilter;
+
+mod hello_world_impl;
+
+/// URL for the chariott service registry
+const CHARIOTT_SERVICE_REGISTRY_URL: &str = "http://0.0.0.0:50000";
```
ashbeitz, line 31: Even though this is a sample, you may want to consider creating a config file for these value.

ashbeitz (commented):
Please take a look at me comments.

service_discovery/proto/build.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft Corporation.
```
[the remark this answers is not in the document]
ladatz: Agreed, seems like this file got duplicated (it's also in the proto_build folder). I will delete this one

service_discovery/proto/chariott/v1/chariott_registry.proto
ladatz: Will change. I had it "DiscoverService" to differentiate from "DiscoverByNamespace" but I agree that Discover matches Register better

service_discovery/core/src/registry_impl.rs
```diff
@@ -0,0 +1,457 @@
...
+
+    /// Unregisters a service by removing it from the registry.
+    ///
+    /// This function registers a service based on a [`UnregisterRequest`]. Returns a
+    /// [`UnregisterResponse`].
+    ///
+    /// # Arguments
+    ///
+    /// * `request` - A [`UnregisterRequest`] wrapped by a [`tonic::Request`].
+    async fn unregister(
```
jorchiu, line 87: This doesn't need to be addressed now, but just wanted to bring this to light. 

Let's say an external provider is registered with Chariott, and there's a consumer that discovers this provider.

I am wondering if this consumer would have enough info from the discovery response to be able to unregister this external provider, or replace this provider in Chariott's registry using the register request?

It might be worth doing some extra validation in the future for unregister() and register().

service_discovery/core/Cargo.toml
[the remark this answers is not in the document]
ladatz: I think I have changed all of my new ones to "_" now, and added license banners

service_discovery/core/src/main.rs
ladatz: I have removed it from all new code, kept it in the existing package name, as well as comments describing the files. Please let me know what you think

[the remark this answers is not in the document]
ladatz: I have created this issue to add configuration in a follow-up change: https://github.com/eclipse-chariott/chariott/issues/144

service_discovery/samples/simple-discovery/consumer/src/main.rs
ladatz: I have created this issue to add configuration in a follow-up change: https://github.com/eclipse-chariott/chariott/issues/144

service_discovery/samples/simple-discovery/provider/src/main.rs
ladatz: I have created this issue to add configuration in a follow-up change: https://github.com/eclipse-chariott/chariott/issues/144. I will think about if we want configs for samples as well.

service_discovery/core/Cargo.toml
ladatz: Issue for adding documentation: https://github.com/eclipse-chariott/chariott/issues/145

service_discovery/core/src/registry_impl.rs
ladatz: Changed

service_discovery/proto/chariott/v1/chariott_registry.proto
ladatz: Upon further thought, I agree that we can leave it simple for now, with no option to "update" a registration once it's been registered. This also brings up some questions that we have about the behavior of Chariott in terms of should we allow multiple running instances of the same version of a service? How will we determine uniqueness? etc. These will come up in future design iterations. For simplicity's sake and until we have a scenario for update, I will remove the "update" part of the register.

service_discovery/proto/build.rs
ladatz: moved

service_discovery/core/src/registry_impl.rs
ladatz: done

Cargo.toml
ladatz: It doesn't look like it. Removed

service_discovery/core/src/registry_impl.rs
ladatz: Agreed. I will create GitHub issues for this, but we need to investigate how we want to handle things like uniqueness and authorization for providers. How do we determine if another provider registering the same version of a service is a unique instance? How do we determine who can register/ unregister a service?

At the start, I see a basic introduction of something that determines if a Chariott service registry client can register or unregister. Over time this can develop by calling into some sort of AuthZ/AuthN service to make the determination or requiring a certificate or unique id. I think it's a great point that we need to investigate more in the future.

service_discovery/proto/build.rs
ladatz: I have removed it from all new code, kept it in the existing package name, as well as comments describing the files. Please let me know what you think

service_discovery/proto/chariott/v1/chariott_registry.proto
ladatz: Added comments for all of the fields in ServiceMetadata

service_discovery/core/src/registry_impl.rs
ladatz: new behavior doesn't allow for overwrite. I am now logging the existing entry

service_discovery/core/src/registry_impl.rs
ladatz: Changed the behavior so that we fail anyone trying to register an existing service. We will revisit this later to determine how we will handle the "update" case

service_discovery/core/src/registry_impl.rs
ladatz: Thanks for the suggestion

service_discovery/core/src/registry_impl.rs
ladatz: updated. I was thinking of it as there are multiple "identifiers" per service (i.e. name, version, etc.) but I think "ServiceIdentifier" can be clearer

service_discovery/samples/simple-discovery/consumer/src/main.rs
ladatz: I have made it a constant at the top of each sample file. As the project expands and there are more and more references to it, we can evaluate adding a common constant file

service_discovery/core/src/registry_impl.rs
```diff
@@ -0,0 +1,457 @@
...
+    }
+
+    /// Unregisters a service by removing it from the registry.
+    ///
+    /// This function registers a service based on a [`UnregisterRequest`]. Returns a
+    /// [`UnregisterResponse`].
+    ///
+    /// # Arguments
+    ///
+    /// * `request` - A [`UnregisterRequest`] wrapped by a [`tonic::Request`].
```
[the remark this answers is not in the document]
ladatz: I will update the comments so that they aren't as verbose

ashbeitz (approved):
Approved

service_discovery/core/src/service_registry_impl.rs
```diff
@@ -0,0 +1,436 @@
...
+    /// * `request` - A [`DiscoverByNamespaceRequest`] wrapped by a [`tonic::Request`].
+    async fn discover_by_namespace(
+        &self,
+        request: Request<DiscoverByNamespaceRequest>,
+    ) -> Result<Response<DiscoverByNamespaceResponse>, Status> {
+        let namespace = request.into_inner().namespace;
+        let service_list: Vec<ServiceMetadata>;
+
+        // This block controls the lifetime of the lock.
+        {
```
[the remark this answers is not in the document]
ladatz: Yes, sorry, this should have been removed.

service_discovery/core/src/service_registry_impl.rs
```diff
@@ -0,0 +1,436 @@
...
+        let service_identifier = ServiceIdentifier {
+            namespace: request_inner.namespace.clone(),
+            name: request_inner.name.clone(),
+            version: request_inner.version,
+        };
+
+        // This block controls the lifetime of the lock.
+        {
+            let lock = self.registry_map.read();
+            match lock.get(&service_identifier) {
```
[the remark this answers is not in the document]
ladatz: Thanks for the suggestion. This was possible by being able to clone the contents of the option. .cloned() "Maps an Option<&T> to an Option<T> by cloning the contents of the option". Updated

service_discovery/samples/simple-discovery/consumer/src/main.rs
```diff
@@ -0,0 +1,70 @@
...
+    });
+
+    // Discover the simple provider service
+    let service_option =
+        service_registry_client.discover(discover_request).await?.into_inner().service;
+    match service_option {
+        Some(service) => {
+            info!("Discovered service {service:?}");
+            if service.communication_kind != EXPECTED_COMMUNICATION_KIND
+                || service.communication_reference != *"hello_world_service.v1.proto"
```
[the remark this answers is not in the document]
ladatz: updated

jorchiu (approved):

service_discovery/core/src/service_registry_impl.rs
```diff
@@ -0,0 +1,420 @@
...
+                    if service_identifier.namespace == namespace {
+                        Some(service_metadata.clone())
+                    } else {
+                        None
+                    }
+                })
+                .collect()
+        };
+        if service_list.is_empty() {
+            Err(Status::not_found(format!("No registrations found for namespace {namespace}")))
```
liviuft, line 140: do we consider this an error or should we just return an empty list?

service_discovery/proto/core/v1/service_registry.proto
```diff
@@ -0,0 +1,107 @@
...
+  // Used by clients to validate that they can communicate with this service. Communication communication_reference
+  // can be a reference to the api specification
+  // example: hello_world_service.v1.proto
+  string communication_reference = 6;
+}
+
+// Status of a registration operation
+enum RegistrationStatus {
+  // An entry did not previously exist in the service registry for this service
+  NEWLY_REGISTERED = 0;
```
liviuft, line 54: do we need this? now that we failed when something is already register

service_discovery/core/src/main.rs
liviuft: we can look at the package names later; maybe open a tracking issue so we do not forget about it

service_discovery/proto/core/v1/service_registry.proto
ladatz: I left it in originally because I anticipate that we will need to add it back at some point, but to keep it more consistent with the other APIs I will remove it.

service_discovery/core/src/service_registry_impl.rs
ladatz: This is a good question. For discovering a single service, we fail if it's not found. But since in this case, the caller would need to consume the list to decide what to do anyway, I will change it to just return an empty list

service_discovery/core/src/service_registry_impl.rs
```diff
@@ -120,52 +111,38 @@ impl ServiceRegistry for ServiceRegistryImpl {
...
-        } else {
-            let discover_by_namespace_response =
-                DiscoverByNamespaceResponse { services: service_list };
-            Ok(Response::new(discover_by_namespace_response))
-        }
+        let service_list: Vec<ServiceMetadata> = {
+            let lock = self.registry_map.read();
+            lock.iter()
+                .filter_map(|(service_identifier, service_metadata)| {
+                    if service_identifier.namespace == namespace {
```
sealemar, line 128: looks like this could be shortened to a more idiomatic [bool.then](https://doc.rust-lang.org/std/primitive.bool.html#method.then):

```rust
(service_identifier.namespace == namespace).then(|| service_metadata.clone())```

ladatz: Thanks, I wasn't aware of this. Updated

sealemar (approved):

PR #152, opened 2023-07-06
Update criterion version to address security vulnerability

ladatz:
## Motivation and Context
There was a security alert for a potential vulnerability with the atty crate. This was referenced in the ess by the criterion crate.

## Description
Update to the latest version of the criterion crate, which doesn't reference atty. Also updated the Cargo.lock file.

### DIFF
--- a/.github/workflows/security-audit.yaml
+++ b/.github/workflows/security-audit.yaml
@@ -20,5 +20,4 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
-      - run: |
-          cargo audit
+      - run: cargo audit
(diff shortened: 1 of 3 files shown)


.github/workflows/security-audit.yaml
```diff
@@ -21,6 +21,4 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - run: |
-          # ignores error due to atty unsound warning
-          # with no solution: TODO remove the ignore
-          cargo audit --ignore RUSTSEC-2021-0145
+          cargo audit
```
devkelley, line 24: suggestion, make the run call a single line, like `run: cargo audit`

devkelley (approved):

PR #174, opened 2023-07-28
Have security audit treat warnings related to a crate as errors too

ladatz:
## Description
In security audit use cargo audit -D warnings to treat crate warnings as errors.

### DIFF
--- a/.github/workflows/security-audit.yaml
+++ b/.github/workflows/security-audit.yaml
@@ -20,4 +20,4 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
-      - run: cargo audit
+      - run: cargo audit -D warnings


.github/workflows/security-audit.yaml
```diff
@@ -20,4 +20,4 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
-      - run: cargo audit
+      - run: cargo audit -D warnings
```
ashbeitz, line 23: Please use:

cargo audit --deny-warnings

See: https://blog.rust-lang.org/inside-rust/2020/01/23/Introducing-cargo-audit-fix-and-more.html

devkelley (approved):

ashbeitz (commented):
Please take a look at my comment.

jorchiu (approved):

.github/workflows/security-audit.yaml
ladatz: I'm unclear on if this is the same as what I have or if it includes yanked crates and the command I have doesn't. Would like to verify if it's the same

ashbeitz: I just chatted with Jordan.  The correct (long form) one to use is:

cargo audit --deny warnings

The one that I gave you earlier was from the 2020 version of cargo audit.

ladatz: Thanks for the suggestion, I think I will leave it as-is, since it's equivalent to what I have, and this repo has "-D warnings" in multiple other places, including for clippy in the rust-ci workflow

ashbeitz (approved):
Approved.

Issue #190, opened 2023-08-30
[PBI] Resolve devskim alerts


ladatz:
## Description
Recently the devskim code scanning was enabled in this repo. Read through and resolve the devskim warnings

## Acceptance criteria

- [x] Resolve http warnings
- [x] Resolve todo warnings
- [x] Resolve any other warnings

ladatz:
The only remaining devskim warning is for code that is copied/ generated from somewhere else.

PR #196, opened 2023-09-27
Address devskim warnings

ladatz:
Addresses #190 

## Motivation and Context
Ignores TLS warning by adding Devskim ignore to http endpoints and localhost. 
## Description
Ignores TLS warning by adding Devskim ignore to http endpoints and localhost. This gets rid of some security warnings that are not truly an issue as they all point to localhost. Most of the "localhost" references are in test code, but some are in our examples, which we don't expect to be used in production. Customers would use their own endpoints for their applications. Also suppress some "TODOs" in code paths that we don't anticipate will have further development on them for now.

Note: the only remaining devskim alerts should be in files that aren't as easy to suppress (i.e. some js files that are marked "generated" and come from somewhere else)

### DIFF
--- a/examples/common/src/chariott/api.rs
+++ b/examples/common/src/chariott/api.rs
@@ -290,7 +290,6 @@ impl<T: ChariottCommunication> Chariott for T {
                     .map(|v| {
                         Value::try_from(v).map_err(|_| Error::new("Could not parse read value."))
                     })
-                    // DevSkim: ignore DS176209 TODO: replace with common::OptionExt after #13
                     .map_or(Ok(None), |r| r.map(Some)),
                 None => Ok(None),
             })
--- a/examples/common/src/chariott/api.rs
+++ b/examples/common/src/chariott/api.rs
@@ -290,7 +290,7 @@ impl<T: ChariottCommunication> Chariott for T {
                     .map(|v| {
                         Value::try_from(v).map_err(|_| Error::new("Could not parse read value."))
                     })
-                    // TODO: replace with common::OptionExt after #13
+                    // DevSkim: ignore DS176209 TODO: replace with common::OptionExt after #13
                     .map_or(Ok(None), |r| r.map(Some)),
                 None => Ok(None),
             })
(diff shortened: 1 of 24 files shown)


examples/common/src/chariott/api.rs
```diff
@@ -290,7 +290,7 @@ impl<T: ChariottCommunication> Chariott for T {
                     .map(|v| {
                         Value::try_from(v).map_err(|_| Error::new("Could not parse read value."))
                     })
-                    // TODO: replace with common::OptionExt after #13
+                    // DevSkim: ignore DS176209 TODO: replace with common::OptionExt after #13
```
ashbeitz, line 293: Perhaps make these two separate comments?

tests/registry-e2e.rs
```diff
@@ -60,7 +60,7 @@ async fn when_provider_registers_notifies_registry_observers() -> anyhow::Result
     let builder = RegistrationBuilder::new(
         "registration.provider.e2e",
         "1.0.0",
-        "http://localhost:7090".parse().unwrap(), // arbitrary url, the provider will never be invoked
+        "http://localhost:7090".parse().unwrap(), // // DevSkim: ignore DS162092 arbitrary url, the provider will never be invoked
```
ashbeitz, line 63: Perhaps make this two separate comments.

ashbeitz (commented):
Some minor comments.  Please take a look.

examples/common/src/chariott/api.rs
ladatz: I actually removed this TODO. I don't think we plan on doing it anyway

wilyle:
As a general comment: why not fix the issues that you can right now instead of making an issue for later? For example, the clippy warning that's being ignored suggests using a struct to encapsulate all of the args which is something that can be done relatively quickly

ladatz:
> As a general comment: why not fix the issues that you can right now instead of making an issue for later? For example, the clippy warning that's being ignored suggests using a struct to encapsulate all of the args which is something that can be done relatively quickly

The short answer is that we have not decided what the future of this code is yet; it may be moved and/ or removed, so I will leave it as-is for now. But I wanted to address most of the devskim alerts for the repo now because we have so much noise it's hard to tell what might be a real issue

devkelley (approved):

ashbeitz (approved):
Approved.

PR #203, opened 2023-10-03
Add GIT_LFS_SKIP_SMUDGE var to failing workflows

devkelley:
## Motivation and Context

The benchmark action runs checkout with LFS. This env var should tell git to checkout without LFS. This will protect the LFS quota that we have.

## Description

The benchmark action runs checkout with LFS. The 'GIT_LFS_SKIP_SMUDGE' env var should tell git to checkout without LFS. This will protect the LFS quota that we have. We do not need the LFS files for these workflows so we are okay using the env variable


ashbeitz (approved):
Approved.

PR #259, opened 2024-02-12
Remove Dog Mode from Chariott repository now that it is in chariott-example-applications repository

### LINKED ISSUE #173: [PBI] Evaluate moving the dogmode demo out of the Chariott repository
## Description
The dogmode demo is a large end-to-end scenario demo. Especially as integration of more components and projects are added, it probably should live outside of the chariott-specific repository.

## Acceptance criteria
- [x] Evaluate where to move it to
- [ ] Move it out
- [ ] Remove all references to it in the chariott repository

ladatz:
Closes #173 

## Motivation and Context

This is solving a separation of concerns problem in which a full scenario example should be separated from the core repository. This way, if people want to clone Chariott core, they do not also need all of the dependencies and the large example scenario. There is now the chariott-example-applications repository which is a more suitable place for it.

## Description

- Delete dogmode related applications and dependencies
- Update documentation


intent_brokering/README.md
```diff
@@ -210,7 +210,9 @@ This walkthrough is described in the [examples kv-app README](./examples/applica
 
 ## How to run the dog mode demo
```
ashbeitz, line 211: Perhaps this section should be removed?

ashbeitz (commented):
One suggestion.  Please take a look.

.dockerignore
```diff
@@ -13,8 +13,6 @@ docs/
 tools/
 
 Cargo.lock
-load_test.sh
```
devkelley, line 16: was this already removed?

ladatz: It was not removed in this PR, but I do not see it in the repository. Since I was removing the other one from this file, I figured I'd clean this up as well.

intent_brokering/README.md
ladatz: I wanted to leave it in for legacy reasons, in case people come here looking for dogmode, but I did updated the link and mentioned that it is hosted in the chariott-example-applications repository

devkelley (approved):

PR #265, opened 2024-02-21
Add multi and arm dockerfiles

devkelley:
## Motivation and Context

This adds a dockerfile that can cross-compile arm64 for docker buildx.

## Description

Dockerfile.service_discovery.multi - can cross-compile either arm64 or amd64 images for service discovery
Dockerfile.service_discovery.arm64 - will build arm64 image with cross compile

### DIFF
--- a/Dockerfile.service_discovery.arm64
+++ b/Dockerfile.service_discovery.arm64
@@ -1,0 +1,56 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+# SPDX-License-Identifier: MIT
+
+ARG RUST_VERSION=1.70
+
+FROM docker.io/library/rust:${RUST_VERSION} AS builder
+
+# Dockerfile for building Eclipse Chariott Service Discovery container
+#
+# This Dockerfile utilizes a two step build process. It builds Chariott
+# Service Discovery with statically linked dependencies (using musl)
+# for a x86_64 architecture.
+
+# Chariott user id
+ARG CHARIOTT_UID=10001
+
+RUN apt update && apt upgrade -y
+RUN apt install -y cmake protobuf-compiler gcc-aarch64-linux-gnu
+
+# unprivileged identity to run Chariott Service Discovery as
+RUN adduser \
+    --disabled-password \
+    --gecos "" \
+    --home "/nonexistent" \
+    --shell "/sbin/nologin" \
+    --no-create-home \
+    --uid "${CHARIOTT_UID}" \
+    chariott
+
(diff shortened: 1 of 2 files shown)


Dockerfile.service_discovery.multi
```diff
@@ -0,0 +1,69 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+# SPDX-License-Identifier: MIT
+
+ARG RUST_VERSION=1.70
+
+FROM --platform=$BUILDPLATFORM docker.io/library/rust:${RUST_VERSION} AS builder
```
ladatz, line 7: Could we add a line or two to the README for the usage of this Dockerfile? I don't think we have any container documentation in this repository yet, but at least if we could show the possible ways to build this one I think it would be a good start

Dockerfile.service_discovery.multi
```diff
@@ -0,0 +1,69 @@
...
+
+# Dockerfile for building Eclipse Chariott Service Discovery container
```
ladatz, line 9: Consider adding a comment here about the multiple target architectures and maybe which ones are currently supported

Dockerfile.service_discovery.multi
```diff
@@ -0,0 +1,69 @@
...
+#
+# This Dockerfile utilizes a two step build process. It builds Chariott
+# Service Discovery with statically linked dependencies (using musl)
+# for a x86_64 architecture.
```
ladatz, line 13: Consider updating this comment with the architecture

Dockerfile.service_discovery.multi
```diff
@@ -0,0 +1,69 @@
...
+
+# Target architecture to cross-compile
+ARG TARGETARCH
```
ladatz, line 16: Should we sanitize this argument?

Dockerfile.service_discovery.arm64
```diff
@@ -0,0 +1,56 @@
...
+
+ARG RUST_VERSION=1.70
+
+FROM docker.io/library/rust:${RUST_VERSION} AS builder
+
+# Dockerfile for building Eclipse Chariott Service Discovery container
+#
+# This Dockerfile utilizes a two step build process. It builds Chariott
+# Service Discovery with statically linked dependencies (using musl)
+# for a x86_64 architecture.
```
ladatz, line 13: Please update this comment with arm64

ladatz (commented):
Please consider my comments

ladatz:
@devkelley also reviewed and approved.

